PR #6, opened 2017-07-02
Refactoring common meta-data fetching from within components to the A…

bryaakov:
…pp component. This also fixed a DisplayNames bug.


matankdr:
LGTM

PR #7, opened 2017-07-03
Fix sshpass copy and path

e-orz:
1. rsync fails if the target directory doesn't exist - fixed
2. In case of a system without ~/bin by default in PATH add it to the custom PATH


e-orz:
LGTM

dudi3001:
LGTM
(for now, as a quick fix, given that this whole module is going to be refactored)

PR #9, opened 2017-07-03
No need to log that index service is terminating

tommy-quitt:
The index service is not terminating anymore therefor, no need to log :-)


tommy-quitt:
Hi @matankdr ,

Apologies for jumping in like that :-)
Anyway, the next line ( indexer.terminate) is remarked thus I guess that the log line should also be remarked.
Had to find something to edit!

Tommy

PR #20, opened 2017-07-05
Readability enhancements, clarified support and environmental URLs...

TRnonodename:
for external audience, updated permid.org PNGs


YoniMataraso:
Dan can you take care of the conflict so we can merge your changes.

PR #26, opened 2017-07-10
Using zCache for UserInfotons

bryaakov:
Some more refactoring is required, as seen to TODOs.
Since it touches too much places, we will do it as a second phase.

### DIFF
--- a/server/cmwell-ws/app/security/AuthCache.scala
+++ b/server/cmwell-ws/app/security/AuthCache.scala
@@ -13,59 +13,46 @@
   * limitations under the License.
   */
 
-
 package security
 
-import java.util.concurrent.TimeUnit
-
 import cmwell.domain.{Everything, FileInfoton}
-import com.google.common.cache.{CacheLoader, CacheBuilder, LoadingCache}
+import cmwell.zcache.L1Cache
+import com.typesafe.scalalogging.LazyLogging
 import logic.CRUDServiceFS
-import play.api.libs.json.{Json, JsValue}
+import play.api.libs.json.{JsValue, Json}
 
 import scala.concurrent.Await
-import scala.concurrent.duration.Duration
-import scala.util.{Success, Try}
 import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.duration._
+import scala.util.Try
 
 /**
- * Created by yaakov on 1/20/15.
- */
-object AuthCache {
+  * Created by yaakov on 1/20/15.
+  */
+object AuthCache extends LazyLogging {
   private val usersFolder = "/meta/auth/users"
   private val rolesFolder = "/meta/auth/roles"
 
+  // TODO Do not Await.result... These should return Future[Option[JsValue]]]
+
   def getUserInfoton(username: String): Option[JsValue] =
-    Try(userInfotonsCache.get(username)).toOption
+    Try(Await.result(usersCache(username), 3.seconds)).toOption.flatten
 
-  def invalidateUserInfoton(username: String) = userInfotonsCache.invalidate(username)
+  def getRole(roleName: String): Option[JsValue] =
+    Try(Await.result(rolesCache(roleName), 3.seconds)).toOption.flatten
 
-  def getRole(roleName: String): Option[JsValue] = {
-    Try(rolesCache.get(roleName)) match {
-      case Success(value) => Some(value)
-      case _ => None
-    }
+  private def getUserFromCas(username: String) = getFromCrudAndExtractJson(s"$usersFolder/$username")
+  private def getRoleFromCas(rolename: String) = getFromCrudAndExtractJson(s"$rolesFolder/$rolename")
+
+  private def getFromCrudAndExtractJson(infotonPath: String) = CRUDServiceFS.getInfoton(infotonPath, None, None).map {
+    case Some(Everything(FileInfoton(_,_,_,_,_,Some(c),_))) =>
+      Some(Json.parse(new String(c.data.get, "UTF-8")))
+    case other =>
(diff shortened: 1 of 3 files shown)


server/cmwell-ws/app/security/AuthCache.scala
```diff
@@ -13,59 +13,46 @@
...
   private val usersFolder = "/meta/auth/users"
   private val rolesFolder = "/meta/auth/roles"
 
-  def getUserInfoton(username: String): Option[JsValue] =
-    Try(userInfotonsCache.get(username)).toOption
+  // TODO Do not Await.result... These should return Future[Option[JsValue]]]
 
-  def invalidateUserInfoton(username: String) = userInfotonsCache.invalidate(username)
+  def getUserInfoton(username: String): Option[JsValue] =
+    Try(Await.result(usersCache(username), 3.seconds)).toOption.flatten
```
hochgi: hard coding `3.seconds` - should be read from conf?

bryaakov: Normally I'd say yes, nothing should be hard-coded... But not this time:

1. This value is how long we wait for CAS to respond, having it different per environment may do some sense, but it's not crucial, becuase we hope any direct read by path should be way less than `3.seconds`, for any environment.
2. This `Await.result` will be very soon gone... So having it in config will just cause noise.

server/cmwell-ws/app/security/AuthCache.scala
```diff
@@ -13,59 +13,46 @@
...
 
-  private val userInfotonsCache: LoadingCache[String, JsValue] =
-    CacheBuilder
-      .newBuilder()
-      .maximumSize(128)
-      .expireAfterWrite(5, TimeUnit.MINUTES)
-      .build(new CacheLoader[String, JsValue] {
-        override def load(key: String) = getFromCrudAndExtractJson(s"$usersFolder/$key")
-      })
+  private def getFromCrudAndExtractJson(infotonPath: String) = CRUDServiceFS.getInfoton(infotonPath, None, None).map {
```
hochgi: you could use `collect` instead of map which discards `other`. It'll return a failed future (with `NoSuchElementException`) on every other case.

server/cmwell-ws/app/security/AuthCache.scala
```diff
@@ -13,59 +13,46 @@
...
-    CacheBuilder
-      .newBuilder()
-      .maximumSize(32)
-      .expireAfterWrite(5, TimeUnit.MINUTES)
-      .build(new CacheLoader[String, JsValue] {
-        override def load(key: String) = getFromCrudAndExtractJson(s"$rolesFolder/$key")
-      })
+  // TODO @inject(ec), do not use Implicits.global
+  private val usersCache = L1Cache.memoize(task = getUserFromCas)(digest = identity, isCachable = _.isDefined)(128, 300)
+  private val rolesCache = L1Cache.memoize(task = getRoleFromCas)(digest = identity, isCachable = _.isDefined)(32,  300)
```
hochgi: Should numbers to fine-tune the cache be taken from conf?

server/cmwell-ws/app/security/AuthCache.scala
```diff
@@ -13,59 +13,46 @@
...
 
-  private val rolesCache: LoadingCache[String, JsValue] =
-    CacheBuilder
-      .newBuilder()
-      .maximumSize(32)
-      .expireAfterWrite(5, TimeUnit.MINUTES)
-      .build(new CacheLoader[String, JsValue] {
-        override def load(key: String) = getFromCrudAndExtractJson(s"$rolesFolder/$key")
-      })
+  // TODO @inject(ec), do not use Implicits.global
```
hochgi: indeed.

server/cmwell-ws/app/security/AuthUtils.scala
```diff
@@ -39,7 +39,6 @@ object AuthUtils {
             val digest2Value = cmwell.util.string.Hash.md5(s"${token.username}:cmwell:$newPw")
             val newUserInfoton = user.as[JsObject] ++ JsObject(Seq("digest"->JsString(digestValue),"digest2"->JsString(digest2Value)))
             CRUDServiceFS.putInfoton(FileInfoton(s"/meta/auth/users/${token.username}", Settings.dataCenter, None, Map.empty[String,Set[FieldValue]], FileContent(newUserInfoton.toString.getBytes, "application/json")))
-            AuthCache.invalidateUserInfoton(token.username)
             true
```
hochgi: Wouldn't it be better to use the `Future` we get from `CRUDServiceFS.putInfoton`, and return a `Future[Boolean]`?
(it already return a `Future[Boolean]`....)

server/cmwell-ws/app/security/AuthCache.scala
bryaakov: Right. But then again, I have to `logger.warn` for cases of None (infoton not preset in CAS) or non FileInfoton which I want to logger.warn about...

server/cmwell-ws/app/security/AuthUtils.scala
bryaakov: will fix it.

server/cmwell-ws/app/security/AuthCache.scala
bryaakov: Good idea, will fix.

server/cmwell-ws/app/security/AuthCache.scala
hochgi: :+1:

server/cmwell-ws/app/security/AuthCache.scala
bryaakov: Will fix on next phase.

PR #28, opened 2017-07-10
set max connections in zookeeper to unlimited

israel:


server/cmwell-cons/app/scripts/templates/zoo.cfg
```diff
@@ -3,4 +3,5 @@ dataDir={{dataDir}}
 clientPort=2181
 initLimit=5
 syncLimit=2
+maxClientCnxns=0
```
e-orz: we tested with 500 which seems more than enough and it wansn't enough.
setting max connections to unlimited will cause the machine socket to be finished

israel:
Happens only in 1 lab env. change is postponed till further investigation

PR #29, opened 2017-07-11
revert zookeeper tcp back to not close connection

e-orz:
without it the client closes the connection before the server's response


e-orz:
This flag is the default for client configuration, no need for comment.
We will never need to change it so make this flag configurable is overkill.

PR #30, opened 2017-07-11
Fixing a typo in AuthCache size

bryaakov:


bryaakov:
Thanks @hochgi for the fast response.

PR #41, opened 2017-07-18
data-post-processor got stuck in data-tools library

matankdr:
Previous code contained a bug if more than 255 sub-streams were present.
DataPostProcessor now returns sorted data using 'buckets' and not sub-streams.
A suject is extract from each triple/quad and the line is appended to its 'bucket' (entry in map).


server/cmwell-data-tools/src/main/scala/cmwell/tools/data/downloader/DataPostProcessor.scala
```diff
@@ -45,11 +45,14 @@ object DataPostProcessor extends DataToolsLogging {
...
-      .fold(Seq.empty[ByteString]){(agg, b) => agg :+ b} // todo: check if endl is needed here
-      .map(concatByteStrings(_, endl))
-      .map(_ ++ endl)
-      .mergeSubstreams
+      .fold(Map.empty[ByteString, ByteStringBuilder]){ (agg, line) =>
+        // aggregate each line according to its subject (i.e., bucket)
+        val subject = GroupChunker.extractSubject(line)
+        val builder = agg.getOrElse(subject, new ByteStringBuilder)
+        builder ++= (line ++ endl)
+        agg + (subject -> builder)
```
e-orz: using immutable map that creates a inner array for *each* line is too time consuming.

PR #60, opened 2017-07-26
optimized file infoton serialization and got rid of wrong warning log

hochgi:
We've seen the following warning in logs:
```
WARN content has different length than expected. expected: [...], actual: [...] for uuid [...]
```
But content was valid. also, `ArrayBuilder` does redundant upcasting and downcasting(via `match`) for value types. we can do better, using a regular array directly.

Also added some minor boy-scouting changes while at it.


server/cmwell-domain/src/main/scala/cmwell/domain/InfotonSerializer.scala
```diff
@@ -163,10 +170,10 @@ object InfotonSerializer extends LazyLogging {
             val content = {
               if (mimeType eq null) None
               else {
-                val content = fileContentBuilder.result()
+                val content = fileContentBuilder//.result()
```
bryaakov: Commented out code... Why not remove `.result()`?

PR #69, opened 2017-07-31
using fileContentCountVerifier rather than checking all zeros

bryaakov:
which is O(1) instead of O(n)


server/cmwell-domain/src/main/scala/cmwell/domain/InfotonSerializer.scala
```diff
@@ -175,7 +175,10 @@ object InfotonSerializer extends LazyLogging {
                 if(fileContentBuildPosition != fileContentLength)
                   logger.warn(s"content has different length than expected. expected: [$fileContentLength], actual: [$fileContentBuildPosition/${content.length}] for uuid [$uuidHint]")
 
-                val contentRes = if((dataPointer ne null) && !content.forall(0.!=)) None else Option(content)
+                val contentRes = {
+                  val hasContent = fileContentCountVerifier > 0
+                  if((dataPointer ne null) && !hasContent) None else Option(content)
```
hochgi: There shouldnt be a case where `dataPointer ne null && hasContent`. So it would be good to have a log line stating this. Will act both as documentation & capture the uuid of a bad case if some bug slipped through...

PR #70, opened 2017-07-31
Iguana tag updates

YoniMataraso:


server/cmwell-bg/src/main/resources/application.conf
```diff
@@ -43,5 +43,6 @@ akka {
   #akka is configured to log in DEBUG level. The actual level is determined by logback
   loglevel = "DEBUG"
   logging-filter = "akka.event.slf4j.Slf4jLoggingFilter"
-  log-config-on-start = off
-}
\ No newline at end of file
+}
+
+cmwell.grid.akka.log-config-on-start = off
```
hochgi: There already a `cmwell.grid`, it should be grouped together. properly nested.

PR #75, opened 2017-08-06
Priority Queue, WS impl.

bryaakov:


server/cmwell-ws/app/controllers/InputHandler.scala
```diff
@@ -62,10 +62,14 @@ class InputHandler @Inject() (ingestPushback: IngestPushback,
    */
   def handlePost(format: String = "") = ingestPushback.async(parse.raw) { implicit req =>
     RequestMonitor.add("in", req.path, req.rawQueryString, req.body.asBytes().fold("")(_.utf8String))
-    val resp = if ("jsonw" == format.toLowerCase) handlePostWrapped(req) -> Future.successful(Seq.empty[(String,String)]) else handlePostRDF(req)
-    resp._2.flatMap { headers =>
-      keepAliveByDrippingNewlines(resp._1,headers)
-    }.recover(errorHandler)
+    if (req.getQueryString("priority").isDefined && !authUtils.isOperationAllowedForUser(security.PriorityWrite, authUtils.extractTokenFrom(req), evenForNonProdEnv = true)) {
```
hochgi: Also, and fix me if I'm wrong here, `authUtils.isOperationAllowedForUser` could be expensive, so order of checks matter. maybe we should document this...

server/cmwell-ws/app/controllers/InputHandler.scala
```diff
@@ -323,6 +327,8 @@ class InputHandler @Inject() (ingestPushback: IngestPushback,
               Future(Ok(Json.obj("success" -> true, "dry-run" -> true)))
             else {
 
+              val isPriorityWrite = req.getQueryString("priority").isDefined
```
hochgi: No point in checking variable again.
memoize as a lazy val (or regular eager val if we always perform this check) at the top of `handlePost`

server/cmwell-ws/app/controllers/Application.scala
```diff
@@ -442,18 +442,19 @@ callback=< [URL] >
       case Failure(e) => asyncErrorHandler(e)
       case Success(request) => {
         val normalizedPath = normalizePath(request.path)
+        val isPriorityWrite = originalRequest.getQueryString("priority").isDefined
```
bryaakov: How `.flatMap(asBoolean).getOrElse(false)` is more concise than `.isDefined` ?
Anyway, "false" won't be a use case. We are going to parse this value as an Integer in phase 2.
For this phase I'm documenting the API to use this query parameter without value, so I hope users won't do `?priority=false` just like they don't do `?with-data=false` because it is simply redundant. Nonetheless, I understand why you wanted to use `asBoolan`, and I would've done it if I wasn't keeping in mind it will soon be an Int. What do you think?

server/cmwell-ws/app/controllers/InputHandler.scala
bryaakov: Makes sense. But then again, I'd have to pass it to handlePostRDF and handlePostWrapped as another Boolean. We already passing the request instance, and `.getQueryString` is O(1) since Play uses a `Map[String,Seq[String]]` for queryString. So there's no need to memoize. And using a lazy val here will actually hurt performance!
I know it feels like I'm not being DRY, but it feels more readable this way.
Let me know what do you think.

server/cmwell-ws/app/controllers/Application.scala
hochgi: Actually, with data does takes a parameter. It's not always applicable though...
But try experiment with `&format=atom&with-data=(trig|nquads|json|etc'...)`

In any case, if it'll be refactored down the road with a new type, it makes even more sense to write it in as few places as possible... Anyway, it doesn't matter that much...

server/cmwell-ws/app/controllers/InputHandler.scala
hochgi: Not sure how lazy val hurts performance here...

PR #79, opened 2017-08-08
Enable stop in stp

matankdr:
* changed `DownloaderStats` and `IngesterStats` to `Flow` instead of `Sink`.
* add hostname in data-tools log messages
* refactor `BufferFillerActor` and simplified its HTTP code


server/cmwell-data-tools/src/main/scala/cmwell/tools/data/downloader/consumer/Downloader.scala
```diff
@@ -124,6 +124,10 @@ object Downloader extends DataToolsLogging with DataToolsConfig{
   def extractTsv(bytes: ByteString) = {
     val arr = bytes.split('\t')
 
+    if (arr.size < 4) {
```
bryaakov: in case one or more infotons don't have indexTime you'll have 3 columns...

e-orz:
The TSV check should be left as is. When getting tsvs using consume there will be always index time and without it's a server failure we want to catch.

PR #80, opened 2017-08-08
added priority topics and adjusted Imp/Indexer Streams to use it

israel:


server/cmwell-bg/src/test/scala/cmwell/bg/test/BGSequentialSpec.scala
```diff
@@ -0,0 +1,162 @@
...
+  var cmwellBGActor:ActorRef = _
+  var dao:Dao = _
+  var irwService:IRWService = _
+  var zStore:ZStore = _
+  var offsetsService:OffsetsService = _
+  var ftsServiceES:FTSServiceNew = _
+  var bgConfig:Config = _
+  var actorSystem:ActorSystem = _
+
+  override def beforeAll = {
```
e-orz: afterAll should also be implemented.

server/cmwell-bg/src/test/scala/cmwell/bg/test/BGSequentialSpec.scala
```diff
@@ -0,0 +1,162 @@
...
+          sortParams = SortParam(("system.indexTime", Asc)),
+          withHistory = false,
+          withDeleted = false
+        )
+
+        Future.sequence(Seq(res1, res2)).map { res =>
+          withClue(res) {
+            val lastOfPriority = res.head.infotons.head.indexTime
+            val firstOfBatch = res.last.infotons.head.indexTime
+            lastOfPriority.get should be >= firstOfBatch.get
```
e-orz: the comparison seems to be the other way around

e-orz:
1. The test is not stable. One time it succeeded and the with the other it finished with
```
[info] - should process priority commands *** FAILED ***
[info]   4351 did not equal 5000 (BGSequentialSpec.scala:142)
```
2. Resilient BG test is now failing (it passed before) with:
```
[info] - should process commands as usual on circumvented BGActor (periodically failing IRWService) after suspending and resuming *** FAILED ***
[info]   (EmptyBox,/cmt/cm/bg-test/circumvented_bg/info0) EmptyBox was empty (BGResilienceSpec.scala:153)
```
4. The test takes 45 seconds to run and this is only one test! We are trying to reduce the test times...
5. This one "This and the next one WON'T be created during upgrade!" still hadn't been fixed yet. The new priority topics (actually also the old ones) won't be created during upgrade.

PR #90, opened 2017-08-17
remove as much -Ds as possible to conf files

e-orz:


server/cmwell-cons/app/pe
```diff
@@ -19,7 +19,7 @@ val pe = LocalHost(dataCenter = "lh",
                    useAuthorization = useAuthorization,
                    withElk = false,
 		   newBg = true,
-		   oldBg = false,
```
israel: why change PE to old?

e-orz: Good Catch! I did it to check the batch changes. I will fix it.

PR #101, opened 2017-08-22
Update response example for x-info (extended in Jaguar release).

nruppin:
Response now includes JSON sent to ES and version in parentheses after index name.


YoniMataraso:
Naomi let's truncate the data that we supply, some of the fields might be considered private, so for company we can use company name address but not more than that.
Can you please update the example again?

PR #103, opened 2017-08-23
Documentation improvements

TRnonodename:
- Optimized all the PNGs and JPGs in /docs/_images
- Added query builder as an example in /docs/examples
- Described and linked to query builder from FileInfoton tutorial


TRnonodename:
OK - fixed 1). Duh, sorry about that. 2) I can't fix and be pedantically HTML. A -- is not a valid character sequence in an HTML comment, so I had put a comment below explaining to others.. Maybe I just move the curl sequence out of the comment and put at the bottom of the page?

PR #106, opened 2017-08-28
Priority topics offsets handling

israel:


server/cmwell-bg/src/main/scala/cmwell/bg/ImpStream.scala
```diff
@@ -307,153 +320,217 @@ class ImpStream(partition:Int, config:Config, irwService:IRWService, zStore: ZSt
...
+          it.remove()
         }
+
         if (prev > lastOffsetPersisted) {
           logger debug s"prev: $prev is greater than lastOffsetPersisted: $lastOffsetPersisted"
           offsetsService.write(s"${streamId}_offset", prev + 1L)
           lastOffsetPersisted = prev
         }
+
+        var prevPriority = lastOffsetPersistedPriority
```
e-orz: please extract method of the logic and use it twice.

PR #108, opened 2017-08-29
priority not mixed with regular

israel:


bryaakov:
Basically, LGTM, but I tried it and when writing an Infoton with `?priority`,
`/proc/bg.md` shows for persist topic **Green/Red**. 
I think this is only a typo in `BGMonitorActor.scala` in lines 107 and 109 - you misplaced the `.p`
They should be `s"imp.${topicPartition.partition()}.p_offset"` and `s"indexer.${topicPartition.partition()}.p_offset"`

So can you add one more commit to this PR?

PR #114, opened 2017-08-30
Consume and Bulk Consume: hanlde a case with lots of infotons with the same index time

e-orz:
1. Consume can now handle infotons with indexTime=0
2. BulkConsume can now handle more than 1000 infotons that have indexTime equals to the "from" section in the token


server/cmwell-ws/app/controllers/Application.scala
```diff
@@ -1130,7 +1130,7 @@ callback=< [URL] >
             case SearchThinResults(total, _, _, results, _) if results.nonEmpty => {
 
               val idxT = results.maxBy(_.indexTime).indexTime //infotons.maxBy(_.indexTime.getOrElse(0L)).indexTime.getOrElse(0L)
-              require(idxT > 0, "all infotons in iteration must have a valid indexTime defined")
+              require(idxT >= 0, "all infotons in iteration must have a valid indexTime defined")
```
hochgi: The whole require is now redundant. Better to delete it...

PR #123, opened 2017-09-03
Allow adding qp to virtual dc infotons

e-orz:
Adding &qp=... to /proc/dc/someId is now allowed.

This is part of adding dc-sync to ability to sync only part of the data. The virtual infoton is used by the monitor and also in dc itself to check what is the last infoton of the desired query.

### DIFF
--- a/server/cmwell-fts/src/main/scala/cmwell/fts/FTSServiceES.scala
+++ b/server/cmwell-fts/src/main/scala/cmwell/fts/FTSServiceES.scala
@@ -1035,7 +1035,7 @@ class FTSServiceES private(classPathConfigFile: String, waitForGreen: Boolean)
     }
   }
 
-  def getLastIndexTimeFor(dc: String, partition: String = defaultPartition)
+  def getLastIndexTimeFor(dc: String, partition: String = defaultPartition, fieldFilters: Option[FieldFilter])
                          (implicit executionContext: ExecutionContext) : Future[Option[Long]] = {
 
     val request = client
@@ -1050,9 +1050,9 @@ class FTSServiceES private(classPathConfigFile: String, waitForGreen: Boolean)
       None,
       Some(MultiFieldFilter(Must, Seq(
         SingleFieldFilter(Must, Equals, "system.dc", Some(dc)),                                 //ONLY DC
-        SingleFieldFilter(MustNot, Contains, "system.parent.parent_hierarchy", Some("/meta/")), //NO META
-        SingleFieldFilter(Must, GreaterThan, "system.lastModified", Some("1970"))
-      ))),
+        SingleFieldFilter(MustNot, Contains, "system.parent.parent_hierarchy", Some("/meta/")) //NO META
+      ) ++ fieldFilters
+      )),
       None
     )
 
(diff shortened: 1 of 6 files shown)


server/cmwell-fts/src/main/scala/cmwell/fts/FTSServiceES.scala
```diff
@@ -1050,9 +1050,9 @@ class FTSServiceES private(classPathConfigFile: String, waitForGreen: Boolean)
       None,
       Some(MultiFieldFilter(Must, Seq(
         SingleFieldFilter(Must, Equals, "system.dc", Some(dc)),                                 //ONLY DC
-        SingleFieldFilter(MustNot, Contains, "system.parent.parent_hierarchy", Some("/meta/")), //NO META
-        SingleFieldFilter(Must, GreaterThan, "system.lastModified", Some("1970"))
-      ))),
+        SingleFieldFilter(MustNot, Contains, "system.parent.parent_hierarchy", Some("/meta/")) //NO META
+      ) ++ fieldFilters
```
hochgi: I think it's best to break down the statement to smaller parts. e.g:
```
val filtersSeq: List[FieldFilter] = ...
Some(MultiFieldFilter(Must,
                      fieldFilters.fold(filtersSeq)(filtersSeq.::))) // same as (_ :: filtersSeq) 
```

server/cmwell-fts/src/main/scala/cmwell/fts/FTSServiceNew.scala
```diff
@@ -1006,9 +1006,9 @@ class FTSServiceNew(config: Config, esClasspathYaml: String) extends FTSServiceO
       None,
       Some(MultiFieldFilter(Must, Seq(
         SingleFieldFilter(Must, Equals, "system.dc", Some(dc)),                                 //ONLY DC
-        SingleFieldFilter(MustNot, Contains, "system.parent.parent_hierarchy", Some("/meta/")), //NO META
-        SingleFieldFilter(Must, GreaterThan, "system.lastModified", Some("1970"))
-      ))),
+        SingleFieldFilter(MustNot, Contains, "system.parent.parent_hierarchy", Some("/meta/")) //NO META
+      ) ++ fieldFilters
```
hochgi: same

server/cmwell-ws/app/controllers/Application.scala
```diff
@@ -311,20 +311,23 @@ callback=< [URL] >
...
-    //    infotonOptionToReply(req,ActiveInfotonGenerator.generateInfoton(req.host, path, new DateTime(),length,offset))
-
-    activeInfotonGenerator
-      .generateInfoton(req.host, path, new DateTime(), length, offset, isRoot, nbg)
-      .flatMap(iOpt => infotonOptionToReply(req, iOpt.map(VirtualInfoton.v2i)))
-  }.recover(asyncErrorHandler).get
+  private def handleGetForActiveInfoton(req: Request[AnyContent], path: String) =
+    getQueryString("qp")(req.queryString)
+      .fold(Success(None): Try[Option[RawFieldFilter]])(FieldFilterParser.parseQueryParams(_).map(Some.apply))
+      .map { qpOpt =>
```
hochgi: dangerous. use `flatMap{ qpOpt => Try { ...`

e-orz: I understand. but...
It's the same pattern as `handleSearch`. Is this a mistake also there?

hochgi: Not nessecarily. `authUtils.isValidatedAs` may fail (e.g: unable to retrieve use infoton), but is not called from search...

Though, I won't argue against being more defensive there too.

hochgi: taking it back... it's actually OK. my bad...

PR #125, opened 2017-09-04
op=stream defaults to 1h scroll ttl

hochgi:


Ivan-Shestakov:
@hochgi  2 questions: 
1) Does it affect other stream types? 
2) Can we override this by providing TTL as query parameter?

hochgi:
@Ivan-Shestakov 
1. No. Just stream
2. Yes, but I added it for debug & benchmarking purposes. End users shouldn't be troubled with it.

PR #126, opened 2017-09-04
bgmonitor does not stop bg actors anymore

israel:


YoniMataraso:
As agreed with @dudi3001, we want this functionality to stay.

PR #144, opened 2017-09-06
Documentation changes for Kingbird release

nruppin:
Update Kingbird release notes.
Add priority flag to _in and POST calls. 
Fix errors in GetNextChunk documentation.


YoniMataraso:
I think the right term we should use for describing the query parameters is query and not call. I think we should revert the changes from query to call.

Issue #142, opened 2017-09-06
/proc/dc virtual infotons should be qp aware


e-orz:

AnnaGrebnev:
@yanpom to test and update status accordingly

PR #152, opened 2017-09-09
query parameter support (qp) in dc monitor and ui

### LINKED ISSUE #138: dc-monitor of syncs with qp
make the dc-monitor be qp-aware.
1.  read the qp from the sync infotons
1. create new dc virtual infotons that accept qp
1. make sure that the monitor won't mix two different qps from the same dc-id
1. stucks
   1. on is stuck and one is working - kill the jvm
   1. one is on horizon and there are more infotons on the source that should be synced - don't kill the jvm

e-orz:
1. DcChecker now checks the sync according to the qp
2. /proc/dc/syncName: can now parse the sync alias name from /meta/sys/dc/syncName and show the correct lastIndexTime taking into account also the qp

Fixes #138 and #142

### DIFF
--- a/server/cmwell-ws/app/logic/CRUDServiceFS.scala
+++ b/server/cmwell-ws/app/logic/CRUDServiceFS.scala
@@ -48,6 +48,7 @@ import scala.concurrent.duration._
 import scala.util.{Failure, Success}
 import javax.inject._
 
+import cmwell.ws.qp.Encoder
 import controllers.NbgToggler
 import ld.cmw.{NbgPassiveFieldTypesCache, ObgPassiveFieldTypesCache}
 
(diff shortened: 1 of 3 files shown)


server/cmwell-ws/app/logic/CRUDServiceFS.scala
```diff
@@ -634,11 +635,14 @@ class CRUDServiceFS @Inject()(tbg: NbgToggler)(implicit ec: ExecutionContext, sy
 
   def getLastIndexTimeFor(dc: String = Settings.dataCenter, nbg: Boolean, fieldFilters: Option[FieldFilter]): Future[Option[VirtualInfoton]] = {
```
hochgi: `fieldFilters` here are taken from `/meta/sys/dc/` w/o any sanitation. So, not tightly related to this PR, but I think it would be best to have a special command/handler to write the `/meta/sys/dc` infotons, and evaluate qp to `DirectFieldKey` where needed.
Also, maybe think of some validation that 2 syncs from same dc with different `qp`s do not intersect (very low priority for now, maybe we can just document that such `qp`s must not intersect)

AnnaGrebnev:
@yanpom to test and update testing status to Done once you finish the testing

PR #171, opened 2017-09-13
Using ApplicationCache to load new UI faster

### LINKED ISSUE #133: Improve loading time of new UI
- Use `<link rel="prefetch" ...` wisely
- Try to load components more lazily

bryaakov:
Fix #133


server/cmwell-spa/data/main/cmwell.appcache
```diff
@@ -0,0 +1,64 @@
+CACHE MANIFEST
+
+# d9c6f43d457a1621cf2f7c2a06b367a0cc7bbddf
```
e-orz: This should be automatically changed.
The suggested change is git_commit_hash+date (full with milliseconds). This way the file will be more or less the same for each git commit (e.g. cm-well version) and also for each build the users will get the new version compiled.
After a short discussion we decided that it will be part of the pakageCmwell sbt task

PR #172, opened 2017-09-14
Adding upload.sh SPA Upload util

bryaakov:


server/cmwell-spa/utils/upload.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+
+tokenFileName=~/dev/uploadTool/token-$1
```
hochgi: Should get from arg. Maybe default to local file named e.g. `token`, and if global default is needed, I think `~/.cmwell/` could be a directory to store such meta data for all our tools.

server/cmwell-spa/utils/upload.sh
```diff
@@ -0,0 +1,51 @@
...
+[ -e "$tokenFileName" ] || touch $tokenFileName -d 2000-01-01
+
+tokenAge=$(($(date +%s) - $(date +%s -r $tokenFileName)))
+if [ $tokenAge -gt 3600 ]
+then
+    token=$(curl -s "http://${1:-localhost:9000}/_login" -u root | jq -r ".token")
```
hochgi: Lets add a doc with requirements, like `jq` should be installed

server/cmwell-spa/utils/upload.sh
```diff
@@ -0,0 +1,51 @@
...
+    echo $token > $tokenFileName
+else
+    token=$(cat $tokenFileName)
+fi
+
+# Create AppCache Manifest according to file list and commit hash
+
+commit=$(git rev-parse HEAD)
+fileList=$(find -type f | grep -vE 'index.html|cmwell.appcache' | sed 's/^\./\/meta\/app\/main/')
```
hochgi: `/meta/app/` should be the default, but lets allow custom apps to any arbitrary path

server/cmwell-spa/utils/upload.sh
```diff
@@ -0,0 +1,51 @@
...
+echo -e "CACHE MANIFEST\n\n# $commit\n\nCACHE:\n$fileList\n\nNETWORK:\n*" > cmwell.appcache
```
hochgi: Continuing the global `~/.cmwell/` dir thought, such files like appcache should be stored there

Ivan-Shestakov:
@bryaakov When this utility would be used?

server/cmwell-spa/utils/upload.sh
```diff
@@ -0,0 +1,51 @@
...
+
+# Upload SPA
+
+for file in $( find -newermt `date +%Y-%m-%d` -type f | grep -v ".idea\|target" | sed 's/^.\///' ); do
```
hochgi: Lets filter all hidden dirs, not just `.idea`

server/cmwell-spa/utils/upload.sh
```diff
@@ -0,0 +1,51 @@
...
+      ico) mime=image/x-icon ;;
+      jpg) mime=image/jpeg ;;
+      gif) mime=image/gif ;;
+      png) mime=image/png ;;
+      svg) mime=image/svg+xml ;;
+      md) mime=text/x-markdown ;;
+      woff2) mime=font/woff2 ;;
+      appcache) mime=text/cache-manifest ;;
+      *)   mime=text/plain ;;
+    esac
```
hochgi: Can we use `/etc/mime.types` for defaults not mention here?

bryaakov:
@Ivan-Shestakov this is only a utility for manually upload an SPA to an existing environment without a cluster upgrade. I use it mainly on localhost when developing new UI features.
Also on Production clusters in case of UI emergencies :)

bryaakov:
@hochgi Thanks for you review. I accepted most of your comments and will soon commit my changes.
In general, this upload.sh was tailored to the default CM-Well UI, and can be easily modified to match one's different requirements, such as ignore other directories.

server/cmwell-spa/utils/upload.sh
bryaakov: :+1:

server/cmwell-spa/utils/upload.sh
bryaakov: :+1:

server/cmwell-spa/utils/upload.sh
bryaakov: :+1:

server/cmwell-spa/utils/upload.sh
bryaakov: I removed this part from this util. see [this commit](https://github.com/thomsonreuters/CM-Well/pull/171/commits/23d340a93217a8992383fecb562e6cb675a9e88f)

server/cmwell-spa/utils/upload.sh
bryaakov: I added some documentation - each one can ignore what ever they like to.

server/cmwell-spa/utils/upload.sh
bryaakov: Good idea. But keeping the static mapping as the first priority. Each laptop can have other etc mimetypes, and we want FileInfotons to be consistent with their mime types, regardless of the uploader machine.

PR #174, opened 2017-09-14
PQ Bloop: Change future work plans

bryaakov:


blps/blp-700-priority-queue.md
```diff
@@ -33,5 +33,4 @@ $ curl "localhost:9000/_in?format=ntriples&priority" --data-binary '<http://exam
 ```
 
 ## Future Work
-* More levels of priority can be introduced. Theoretically, the `priority` query parameter can accept an Integer value from 0 (which is equivalent to not providing the query parameter at all) to MAX_INT. But we will merely add one more, i.e. the values of 1 or 2 can be used. 1 means medium priority and 2 means high priority.
-* The change above can be reflected in Authorization as well, i.e. the Priority Role will accept a value of max priority allowed for user. This part is optional, we can decide any user with the Priority role can use any value of priority query parameter.
+* One or a few more (up to a fixed small number) priorites can be introduced. For example "idle queue" which is intended for Admin operations, which will be prioritzed higher than normal but lower than the Priority priority.
```
YoniMataraso: I would change the the lower than the Priority priority to something like lower than the high Priority queue.

PR #187, opened 2017-09-24
Faster redirection to Old UI when /proc/* is requested

bryaakov:


server/cmwell-spa/data/main/scripts/main.js
```diff
@@ -1,4 +1,7 @@
 (function(){
+
+    
+if(location.pathname.indexOf('/proc')>-1) location.href+='?old-ui';
```
e-orz: even if an infoton will contain /proc it will fall back to the old-ui (e.g. /example.org/process/....)
You should search that it will *start* with /proc/ (note the trailing / for /process.com/...)
Also, don't forget plain /proc.

yanpom:
Tested

PR #188, opened 2017-09-24
Better exception handling in LoginHandler

### LINKED ISSUE #182: Incorrect credentials Login API (basic authentication) making server to hang
When submitting the following request to the login API the server will hang and will not respond in an reasonable time/response.
curl 'http://localhost:9000/_login' -H 'Authorization:Basic testUser:testUser'

The expected behavior would be not Authenticated response(HTTP 401)

bryaakov:
Fixes #182

### DIFF
--- a/server/cmwell-ws/app/security/httpauth/BasicHttpAuthentication.scala
+++ b/server/cmwell-ws/app/security/httpauth/BasicHttpAuthentication.scala
@@ -16,15 +16,13 @@
 
 package security.httpauth
 
-import org.apache.commons.codec.binary.Base64
-
-import scala.util.{Failure, Success, Try}
+import cmwell.util.string.Base64.decodeBase64String
 
 trait BasicHttpAuthentication {
   private val authTypePrefix = "Basic "
 
   def decodeBasicAuth(auth: String): (String, String) = {
-    val userColonPass = new String(new Base64().decode(auth.drop(authTypePrefix.length)), "UTF-8")
+    val userColonPass = decodeBase64String(auth.drop(authTypePrefix.length), "UTF-8")
     val split = userColonPass.split(":", 2)
     if (split.length != 2)
       throw new IllegalArgumentException(s"`$auth` is not a valid Basic header")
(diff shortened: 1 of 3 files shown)


bryaakov:
@hochgi, please review my changes in `package.scala`

server/cmwell-ws/app/security/httpauth/BasicHttpAuthentication.scala
```diff
@@ -23,15 +23,12 @@ import scala.util.{Failure, Success, Try}
...
-      case Success(userColonPass) =>
-        val split = userColonPass.split(":", 2)
-        if (split.length != 2)
-          Failure(new IllegalArgumentException(s"$auth is not a valid Basic header"))
-        else
-          Success(split(0)->split(1))
-      case Failure(_) => Failure(new IllegalArgumentException(s"$auth is not a valid Basic header"))
-    }
+  def decodeBasicAuth(auth: String): (String, String) = {
+    val userColonPass = new String(new Base64().decode(auth.drop(authTypePrefix.length)), "UTF-8")
```
hochgi: Use `cmwell.util.string.Base64`, and it might be worth validating before decoding, with: `Base64.isBase64(...)`.
No need for `new` instance

server/cmwell-ws/app/security/httpauth/BasicHttpAuthentication.scala
```diff
@@ -23,15 +23,12 @@ import scala.util.{Failure, Success, Try}
...
+    val split = userColonPass.split(":", 2)
```
hochgi: Might be better to use `cmwell.util.string.splitAtNoSep`

bryaakov: I disagree. I need to validate the input was in form of `base64(user:pass)` (i.e. after invoking decode, that the split returned an exact length of 2 elements). `splitAtNoSep` is too forgiving (e.g. it returns `(str,"")` if separator does not exist in input. I don't want that behavior here, or I will return 401 even for 400 cases.

PR #195, opened 2017-09-26
DcSync: Properly take the http body size from configuration

### LINKED ISSUE #194: dc sync falls back to slow consume api
DC sync falls back to slow consume api whenever the chunked response exceeds 8MB (akka default).

e-orz:
It seems that using ConnectionPoolSettings withing the newHostConnectionPool takes the fallback config only from reference.conf and not application.conf
After the configuration changes the config was moved to application.conf.
The solution is to manually take the config.

Fixes #194


Ivan-Shestakov:
Validated by reviewing dc sync log and finding instances where dc sync retrieved chunks larger than default 8MByte.

PR #197, opened 2017-09-26
AuthCache: Retring on 404 from CAS

### LINKED ISSUE #196: AuthCache: Retry on 404 from CAS

bryaakov:
Fixes #196


YoniMataraso:
@hochgi 
AFAIK the retry mechanism will only fail in cases that an error returned from the call?
In this case we are trying to put a temporary solution where Cassandra might not respond correctly as we saw in our production environment.

The existing retry mechanism works agains failures and not in cases where no data was found, this is only being added where it hurts us most.

hochgi:
@YoniMataraso it does retries when no data was found. as stated in previous comment as "404". and this is a retry on both `path` queries (to get `uuid`), and again on `uuid` queries (to get the actual `Infoton`)

PR #198, opened 2017-09-27
ignoring known sys fields that has no effect in ingests

hochgi:


AnnaGrebnev:
@Ivan-Shestakov to test it,  can't add you as assignee

PR #220, opened 2017-10-02
fixed bug in OW processing where re-processing of already persisted while not indexed yet OWC, was indexed with current:false

israel:
fixed bug in OW processing where re-processing of already persisted while not indexed yet OWC, was indexed with current:false.
added deleteInfotons method to FTSServiceNew

### DIFF
--- a/server/cmwell-bg/src/main/scala/cmwell/bg/CMWellBGActor.scala
+++ b/server/cmwell-bg/src/main/scala/cmwell/bg/CMWellBGActor.scala
@@ -55,6 +55,8 @@ class CMWellBGActor(partition:Int, config:Config, irwService:IRWService, ftsServ
   var impStream:ImpStream = null
   var indexerStream:IndexerStream = null
   val waitAfter503 = config.getInt("cmwell.bg.waitAfter503")
+  val impOn = config.getBoolean("cmwell.bg.ImpOn")
+  val indexerOn = config.getBoolean("cmwell.bg.IndexerOn")
 
   // Metrics
   val jmxReporter = JmxReporter.forRegistry(metricRegistry).build()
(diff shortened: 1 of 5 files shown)


server/cmwell-bg/src/main/scala/cmwell/bg/CMWellBGActor.scala
```diff
@@ -177,19 +179,23 @@ class CMWellBGActor(partition:Int, config:Config, irwService:IRWService, ftsServ
   }
 
   private def startImp = {
-    if(impStream == null) {
-      logger info "starting ImpStream"
-      impStream = new ImpStream(partition, config, irwService, zStore, ftsService, offsetsService, self)
-    } else
-      logger warn "requested to start Imp Stream but it is already running. doing nothing."
+    if(impOn) {
+      if (impStream == null) {
```
hochgi: can't we get along with a simple `lazy val impStream = new ImpStream(...)` instead of `var impStream: ImpStream = null` with this check?
It can potentially be dangerous with multi-threaded execution...

server/cmwell-bg/src/main/scala/cmwell/bg/CMWellBGActor.scala
```diff
@@ -177,19 +179,23 @@ class CMWellBGActor(partition:Int, config:Config, irwService:IRWService, ftsServ
...
+        logger info "starting ImpStream"
+        impStream = new ImpStream(partition, config, irwService, zStore, ftsService, offsetsService, self)
+      } else
+        logger warn "requested to start Imp Stream but it is already running. doing nothing."
+    }
   }
 
   private def startIndexer = {
```
hochgi: same for `startIndexer` better have a `lazy val` instead `var` + `private def` with nullity check

server/cmwell-bg/src/test/scala/cmwell/bg/test/CmwellBGSpec.scala
```diff
@@ -466,6 +467,58 @@ class CmwellBGSpec extends AsyncFunSpec with BeforeAndAfterAll with Matchers wit
...
+
+      // send them all
+      val sendEm = Future.traverse(pRecords)(sendToKafkaProducer)
+
+      sendEm.flatMap { _ =>
+        scheduleFuture(5.seconds) {
+          ftsServiceES.deleteInfotons(infotons).flatMap { _ =>
+            scheduleFuture(5.seconds) {
+              val sendAgain = Future.traverse(pRecords)(sendToKafkaProducer)
+              scheduleFuture(5.seconds) {
```
hochgi: all the scheduling can be replaced with `skipCheck`s (see integration tests `Helpers` trait). it will:
- perform faster (5 seconds are probably too much, and we need 3 waits)
- be safer, and won't fail in case it takes longer than 5 seconds (due to e.g: gc)

israel: will be done as part of improving all BG tests

server/cmwell-bg/src/main/scala/cmwell/bg/CMWellBGActor.scala
israel: this null is used in other places

server/cmwell-bg/src/main/scala/cmwell/bg/CMWellBGActor.scala
israel: same as above

PR #224, opened 2017-10-15
SpHandler Fix for detecting data issues feature

bryaakov:
This fixes the issue of false positive error of "Fetching [some path] had no results, this might be a data issue", e.g. with yg of no results.


bryaakov:
@yanpom to test this change, please try posting to `_sp` something like this:
```
PATHS
/no-such-infoton?yg=<somePredicate

SPARQL
SELECT * WHERE { ?s ?p ?o }
```
_(should be 500 before the fix and 200 afterwards)._

yanpom:
Tested

PR #226, opened 2017-10-15
prepare machines: configure the network of GridSubDiv only on the pre…

e-orz:
…pared machine and not the whole cluster


e-orz:
Fixes #226

PR #248, opened 2017-10-20
STP - adding logs+flow fixes+config fix (second try)

e-orz:
1. adding logs to STP stand alone to print the reason the stream finished
2. fix the alsoToMat in BufferFillerActor to propagate the stream end to the upstream
3. fix connection pools to not just use the library defaults but also the application.conf values


hochgi:
@e-orz What's changed from #242 ?

e-orz:
@hochgi  Nothing just Jenknins couldn't compile the old one so I created a new one rebased on the current master - and it worked!

PR #281, opened 2017-11-08
Authcache GraceDeg fix

### LINKED ISSUE #276: UserAuthentication - Ingesting user with bad syntax causes cache loading to fail
Ingest user with bad syntax (corrupted json)
Try to login with root:root
Login fails because one of the users syntax is invalid
The fact that one of the users corrupted, should not fail the entire AuthCache loading

2017-11-02 12:23:45,865 pplication-akka.actor.default-dispatcher-32] INFO  security.AuthCache                   - AuthCache failed to load because Unexpected character ('+' (code 43)): was expecting comma to separate Object entries
 at [Source: (String)"{"paths":[{"id":"/","sign":"+","recursive":true,"permissions":"r"},{"id":"/test.permid.org/userTest/luba04","sign":"+","recursive":true,"permissions":"rw"}],"digest":"$2a$10$o1WYcW/UbausG7PdjF9Sze8rR/LoLYtyYPMVL3DhY4c42dSC56rOK","digest2":"" + hash + ""}"; line: 1, column: 244]

bryaakov:
- In case when Search Future will be failing this will work as it is today. But in case it will only return after 10000 millis (i.e. fts timeout), well, current implementation won't fallback to direct read. Added `timeoutFuture` to address this issue.
- Fix #276 with a simple `Try` wrapped `Json.parse` and added relevant log msgs.

### DIFF
--- a/server/cmwell-ws/app/security/AuthCache.scala
+++ b/server/cmwell-ws/app/security/AuthCache.scala
@@ -30,19 +30,35 @@ import scala.concurrent.{Await, ExecutionContext, Future}
 @Singleton
 class AuthCache @Inject()(crudServiceFS: CRUDServiceFS)(implicit ec: ExecutionContext) extends LazyLogging {
   // TODO Do not Await.result... These should return Future[Option[JsValue]]]
-  def getRole(roleName: String, nbg: Boolean): Option[JsValue] = Await.result(data(nbg).getAndUpdateIfNeeded.map(_.roles.get(roleName)).recoverWith { case _ =>
-    logger.warn(s"AuthCache Graceful Degradation: Search failed! Trying direct read for Role($roleName):")
-    getFromCrudAndExtractJson(s"/meta/auth/roles/$roleName", nbg)
-  }, 5.seconds)
+  def getRole(roleName: String, nbg: Boolean): Option[JsValue] = {
+    val esDependingFut = data(nbg).getAndUpdateIfNeeded.map(_.roles.get(roleName))
+    Await.result(
+      cmwell.util.concurrent.timeoutOptionFuture(esDependingFut, 3.seconds).flatMap {
+        case Some(roleInfotonOpt) => Future.successful(roleInfotonOpt)
+        case None =>
+          logger.warn(s"AuthCache Graceful Degradation: Search failed! Trying direct read for Role($roleName):")
+          getFromCasAndExtractJson(s"/meta/auth/roles/$roleName", nbg)
+      }
+      , 6.seconds
+    )
+  }
 
-  def getUserInfoton(userName: String, nbg: Boolean): Option[JsValue] = Await.result(data(nbg).getAndUpdateIfNeeded.map(_.users.get(userName)).recoverWith { case _ =>
-    logger.warn(s"AuthCache Graceful Degradation: Search failed! Trying direct read for User($userName):")
-    getFromCrudAndExtractJson(s"/meta/auth/users/$userName", nbg)
-  }, 5.seconds)
+  def getUserInfoton(userName: String, nbg: Boolean): Option[JsValue] = {
+    val esDependingFut = data(nbg).getAndUpdateIfNeeded.map(_.users.get(userName))
+    Await.result(
+      cmwell.util.concurrent.timeoutOptionFuture(esDependingFut, 3.seconds).flatMap {
+        case Some(userInfotonOpt) => Future.successful(userInfotonOpt)
+        case None =>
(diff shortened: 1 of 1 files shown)


server/cmwell-ws/app/security/AuthCache.scala
```diff
@@ -26,23 +26,40 @@ import play.api.libs.json.{JsValue, Json}
...
 @Singleton
 class AuthCache @Inject()(crudServiceFS: CRUDServiceFS)(implicit ec: ExecutionContext) extends LazyLogging {
   // TODO Do not Await.result... These should return Future[Option[JsValue]]]
-  def getRole(roleName: String, nbg: Boolean): Option[JsValue] = Await.result(data(nbg).getAndUpdateIfNeeded.map(_.roles.get(roleName)).recoverWith { case _ =>
-    logger.warn(s"AuthCache Graceful Degradation: Search failed! Trying direct read for Role($roleName):")
-    getFromCrudAndExtractJson(s"/meta/auth/roles/$roleName", nbg)
-  }, 5.seconds)
+  def getRole(roleName: String, nbg: Boolean): Option[JsValue] = {
+    val esDependingFut = data(nbg).getAndUpdateIfNeeded.map(_.roles.get(roleName))
+    Await.result(
```
hochgi: Are we still using Await?
can we format it as:
```
val f = ...
Await.result(f,6.seconds)
```
instead of:
```
Await.result(
  ...
  , 6.seconds
)
```

server/cmwell-ws/app/security/AuthCache.scala
```diff
@@ -26,23 +26,40 @@ import play.api.libs.json.{JsValue, Json}
...
+      }
+      , 6.seconds
+    )
+  }
 
-  def getUserInfoton(userName: String, nbg: Boolean): Option[JsValue] = Await.result(data(nbg).getAndUpdateIfNeeded.map(_.users.get(userName)).recoverWith { case _ =>
-    logger.warn(s"AuthCache Graceful Degradation: Search failed! Trying direct read for User($userName):")
-    getFromCrudAndExtractJson(s"/meta/auth/users/$userName", nbg)
-  }, 5.seconds)
+  def getUserInfoton(userName: String, nbg: Boolean): Option[JsValue] = {
```
hochgi: can we factor out commonalities of `getRole` and `getUserInfoton`?
code is almost identical.

server/cmwell-ws/app/security/AuthCache.scala
```diff
@@ -83,9 +100,12 @@ class AuthCache @Inject()(crudServiceFS: CRUDServiceFS)(implicit ec: ExecutionCo
 
   private def extractPayload(infoton: Infoton): Option[JsValue] = infoton match {
     case FileInfoton(_, _, _, _, _, Some(c), _) =>
-      Some(Json.parse(new String(c.data.get, "UTF-8")))
+      val jsValOpt = Try(Json.parse(new String(c.data.get, "UTF-8"))).toOption
```
hochgi: 1. `Json.parse` also takes an `Array[Byte]`, So no need to `new String(...)`
2. if`c.data.get` might fail, we should catch it either in the pattern match case, or deal with `Try` failures, instead of hiding the exception, and logging (erroneously) that we got an invalid json. pattern match can be:
```
case FileInfoton(_, _, _, _, _, Some(FileContent(Some(c),_,_,_)), _) =>
```
Also, is it possible we get a `dataPointer` instead the data?
Shouldn't we handle such a case?

server/cmwell-ws/app/security/AuthCache.scala
bryaakov: Yes, I kept that `TODO` to some other time. And yes, I can do that `val f`.

server/cmwell-ws/app/security/AuthCache.scala
bryaakov: I'll give it a shot.

server/cmwell-ws/app/security/AuthCache.scala
bryaakov: 1. Ok! I will remove the `new String`.
2. I'll add a case.
3. No, IRW is responsible for populating data in case when `dataPointer` is in use, and logically we will never get an Infoton without content. Also - in this case I really hope no UserInfoton will reach zStore threshold, LOL.

hochgi: Reg (3), then it shouldn't be leaked.
We should open another issue to fix `FileContent` class.

hochgi: ref #284

PR #294, opened 2017-11-13
Doc update user management

nruppin:
Added 4 new files related to user management and its APIs.
Updated TOC files related to user management.


docs/DevGuide.ManagingUsers.md
```diff
@@ -0,0 +1,137 @@
...
+>**Notes**: 
+>* The JSON body does not include the username, as the username is "stored" as the FileInfoton's name.
+>* To call the **_login** API, the user will need either a **digest** or a **digest2** value defined.
+
+Here is an example of a user FileInfoton's JSON body:
+
+    {
+      "paths": [
+	      {
+		      "id": "\/",
```
bryaakov: no need for escaping here. correct line should be `"id": "/",`

docs/DevGuide.ManagingUsers.md
```diff
@@ -0,0 +1,137 @@
...
+		      "sign": "+",
+		      "recursive": true,
+		      "permissions": "r"
+	      },
+	      {
+		      "id": "\/test.example.org",
```
bryaakov: same as above. correct line should be `"id": "/test.example.org",`

docs/DevGuide.ManagingUsers.md
```diff
@@ -0,0 +1,137 @@
...
+
+1. Login to CM-Well with an admin user. A token with admin permissions is returned in the response.
+1. Create the JSON body with role paths and permissions as required. See [paths field](#tableFields).
+1. Upload the JSON body you created to a FileInfoton named <rolename>, under the folder ```/meta/auth/roles```. For example:
+
+    ```curl cmwell/meta/auth/roles/Editor -H "X-CM-Well-Type:File" -H "Content-Type:application/json" -H "X-CM-Well-Token:<AdminToken>" --data-binary @roleDetails.json```
+
+1. In order for the new role information to take effect, refresh the CM-Well authentication cache using the following command:
+
+    ```\<cm-well-host\>/_auth?op=invalidate-cache -H "X-CM-Well-Token:<AdminToken>"```
```
bryaakov: no need to escape `<` and `>` with a backslash, in case when whole thing is wrapped by ` ``` `s. Just GitHub Markdown things... i.e. correct line is ```<cm-well-host>/_auth?op=invalidate-cache -H "X-CM-Well-Token:<AdminToken>"```

docs/API.Auth.InvalidateCache.md
```diff
@@ -0,0 +1,32 @@
+# Function: *Invalidate Cache* #
+
+## Description ##
+
+CM-Well manages a cache with authentication information for better performance. If you've made a change to authentication information such as creating a new user, creating a new role or changing a password, and you want the change to take immediate effect, you will need to refresh this cache. You do this by calling the **invalidate-cache** API.
```
bryaakov: CM-Well manages a cache with <s>authentication</s> **authorization** information for better performance. If you've made a change to <s>authentication</s> **authorization** information such as ...

docs/API.Auth.InvalidateCache.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Code Example ##
+
+### Call ###
+
+    <CMWellHost>/_auth?op=invalidate-cache -H "X-CM-Well-Token:<AdminToken>"
+
+### Results ###
+
+    {"sucess":true}
```
bryaakov: as you wrote on another place - I think it is best if also mentioned here - on rare cases response might be "success: false". In that case, just try again. (this is because the cache might be just updating itself in the exact time of the invalidation request, so it is "locked" and will report the operation as unsuccessful).

docs/API.Auth.GeneratePassword.md
```diff
@@ -0,0 +1,30 @@
+# Function: *Generate Password* #
+
+## Description ##
+
+When creating a new CM-Well user, you can use the **generate-password** API to generate a random password and its basic-encoded value. This password can be changed later if required (see [Change Password API](API.Auth.ChangePassword.md)).
```
bryaakov: rather than "basic-encoded" I'd say "encrypted"

docs/API.Auth.GeneratePassword.md
```diff
@@ -0,0 +1,30 @@
...
+
+>**Note:** You will need an access token with admin permissions to call this API.
```
bryaakov: Actually, since it's not affecting anything in CM-Well, there's no need for an Admin token here. Anyone can generate-password as much as they like - only when used later in a UserInfoton it has some meaning. This API is just a "helper method" to invoke the bcrypt algorithm on a random value with CM-Well Cluster's private key.

docs/API.Auth.GeneratePassword.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Syntax ##
+
+**URL:** \<CMWellHost\>/_auth?op=generate-password -H "X-CM-Well-Token:\<AdminToken\>"
```
bryaakov: as explained above - no need for token header.

docs/API.Auth.GeneratePassword.md
```diff
@@ -0,0 +1,30 @@
...
+**REST verb:** GET
+**Mandatory parameters:** Admin-permissions token in the **X-CM-Well-Token** header
```
bryaakov: again - no need for token (just a reminder to remove this line as well)

docs/API.Auth.GeneratePassword.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Code Example ##
+
+### Call ###
+
+    curl <cm-well-host>/_auth?op=generate-password -H "X-CM-Well-Token:<AdminToken>"
```
bryaakov: just "curl <cm-well-host>/_auth?op=generate-password".

@nruppin sorry if I was misleading you about this specific API. I might have said once "all _auth operations required an Admin token" which was clearly a mistake - this API is an exception.

docs/API.Auth.ChangePassword.md
```diff
@@ -0,0 +1,37 @@
...
+Parameter | Description 
+:---------|:-------------
+current   | The user's current password
+new       | The user's new password
+
+## Code Example ##
+
+### Call ###
+
+    <CMWellHost>/_auth?op=change-password&current=$2a$10$7AnXsjks.IZXTbpRiAGN4OQItwiz4sgxM49lvTiCjWgOhbbOQkg2m&new=12345678 -H "X-CM-Well-Token:<AdminToken>"
```
bryaakov: same as above. for the "current" value - use the plain-text password, not its encrypted value

PR #301, opened 2017-11-17
Error 422; piped yg/gqp; range parameters supported for xg/yg

nruppin:
Add notes about 422 error when deleting non-existent sub-graphs.
Remove notes about <,>,<<,>> not being supported for xg/yg.
Add sections about piped expressions for yg/gqp.

### DIFF
--- a/docs/API.ReturnCodes.md
+++ b/docs/API.ReturnCodes.md
@@ -9,5 +9,6 @@ Return Code | Description
 204 | No more results
 404 | Infoton not found
 401 | Unauthorized user, or password does not match user name
+422 | The sub-graph you requested to delete does not exist
 503 | Server is busy
 5xx | Access problem
--- a/docs/API.ReturnCodes.md
+++ b/docs/API.ReturnCodes.md
@@ -9,6 +9,6 @@ Return Code | Description
 204 | No more results
 404 | Infoton not found
 401 | Unauthorized user, or password does not match user name
-422 | The sub-graph you requested to delete does not exist
+422 | The request has no effect on CM-Well data
 503 | Server is busy
 5xx | Access problem
--- a/docs/API.TraversingOutboundAndInboundLinksWithXgAndYg.md
+++ b/docs/API.TraversingOutboundAndInboundLinksWithXgAndYg.md
@@ -14,11 +14,13 @@
 		* [The Field Wildcard](#hdr8)
 		* [Field Filters](#hdr9)
 * [Traversing Outbound and Inbound Links with the **yg** Flag](#hdr10)
-* [Using the **yg-chunk-size** Parameter](#hdr11)
+    * [Implementing Multiple Expansion Paths with the Pipe Operator](hdrYgPipe)
+    * [Using the **yg-chunk-size** Parameter](#hdr11)
 * [Using **xg** and **yg** Together](#hdr12)
 	* [Ghost Skips](#hdr13)
-* [Usage Scenario Using Outbound and Inbound Links](#hdr14)
+* [Usage Scenario with Outbound and Inbound Links](#hdr14)
 * [Using **gqp** to Filter by Inbound/Outbound Relationships](#hdr15)
+    * [Implementing Multiple Filter Expressions with the Pipe Operator](#hdr16)
 
 ----------
 
@@ -255,9 +257,7 @@ You can add field filters to constrain the additional infotons retrieved by the 
 
     xg=<outboundExpansion>[field filter]
 
-The syntax of the field filter is the same as for the [qp parameter](API.FieldConditionSyntax.md), except for the following differences:
-* Only the partial match and exact match operators (: and ::) are supported for the **xg** flag. (The fuzzy match operator ~ is not supported for **xg**.)
-* Range queries (<,>,<<,>>) are not supported for **xg**.
+>**Note:** The syntax of the field filter is the same as for the [qp parameter](API.FieldConditionSyntax.md), except that the fuzzy match operator ~ is not supported for **xg** queries.
 
 For example:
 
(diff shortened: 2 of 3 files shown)


docs/API.ReturnCodes.md
```diff
@@ -9,5 +9,6 @@ Return Code | Description
 204 | No more results
 404 | Infoton not found
 401 | Unauthorized user, or password does not match user name
+422 | The sub-graph you requested to delete does not exist
```
hochgi: It's not about non existent sub graphs. Every request that is semantically correct, but has no effect or meaning, should result in 422. Another example, is ingesting something like:
```
<http://my.domain.com/some/path> <cmwell://meta/sys#path> "/my.domain.com/some/path" .
```

docs/API.TraversingOutboundAndInboundLinksWithXgAndYg.md
```diff
@@ -274,15 +274,25 @@ Much of the syntax for the **yg** flag is the same as for the **xg** flag, so pl
...
 This query retrieves all infotons that point to the matched infotons through their immediate parents field, and also all infotons that point to the level 1 inbound links through their immediate parents field. In other words, it retrieves the child companies of the child companies of the matched infotons. 
 
+<a name="hdrYgPipe"></a>
+### Implementing Multiple Expansion Paths with the Pipe Operator ###
+
+If you want to expand along several paths using a single query, you can add several expansion expressions separated by the pipe ('|') operator.
+
+For example, the following query searches for organizations whose name contains "Marriott", and also retrieves Quote and Instrument infotons that are issues by those organizations. Note the use of the '|' operator to expand along two different inbound links.
+
+    <cm-well-host>/permid.org?op=search&qp=CommonName.mdaas:Marriott&yg=<isQuoteOf.mdaas|<isIssuedBy.mdaas
```
hochgi: This example is not a good usecase for pipes, as it can be wrtten using regular commas instead. Let's change to 2 path differentiating by direction. E.g: `<field.ns|>anotherField.ns`, or something with more meat in it.

docs/API.TraversingOutboundAndInboundLinksWithXgAndYg.md
```diff
@@ -349,12 +359,21 @@ There are ways to address this using **yg** and SPARQL queries, but they have li
...
+>* The [ghost skips](#hdr13) behavior applies to **qgp** as well as **yg**.
+
+<a name="hdr16"></a>
+## Implementing Multiple Filter Expressions with the Pipe Operator ##
+
+As for the **yg** query, you can add several filter expressions to a single **gqp** query, separated by the pipe ('|') operator. However, the behavior of the complex expression is different than for **yg**. When piped expressions are used with **gqp**, CM-Well applies "or" logic among them. This means that in order to match the filter, an infoton only needs to match *one* of the piped expressions.
+
+For example, the following query retrieves infotons of people who live *either* in New York *or* in Chicago.
+
+    <cm-well-host>/?op=search&qp=type.rdf:Person&gqp=<addressOfPerson>physicalAddress[city::New%20York]|<addressOfPerson>physicalAddress[city::Chicago]
```
hochgi: Again, not very good example as the filter could be `*city::Chicago,*city::New%20York`, which is preferable.

docs/API.TraversingOutboundAndInboundLinksWithXgAndYg.md
nruppin: Gilad, please suggest authentic code examples for these two cases. I have a hard time coming up with them myself (as you can see).

hochgi: /permid.org might not be the best data to demonstrate such features, as it is highly denormalized.
How bout an example like:

Person --placeOfBirth--> Location
Person <--hasTenant-- Residence --locatedIn--> Location

where we try to find all the location a person been at.
the yg query could be something like:

yg=<hasTenant>locatedIn|>placeOfBirth

WDYT?

nruppin: Not sure I understand this completely. Does it look for all locations a person has been at, or all people who have lived at a certain location? Isn't the initial query on Residence infotons (not Person infotons)? Can you please send me the entire query?

Also, can you send me a good example for gqp as well (your comment below)?

hochgi: Usage is the same. when you want to start on expansion with '>', and one with '<'.
regarding the suggested example, let's use some geonames data for locations:
http://sws.geonames.org/4140704/about.rdf
http://sws.geonames.org/5122525/about.rdf
and add some properties of our own, which results in [this](https://gist.github.com/hochgi/05af93055c555ae68f96189e957043b2)

Now, when the above is ingested to CM-Well, you could do a yg query as follows:
```
<CM-Well HOST>/example.org/Individuals/Donald_Trump?yg=<hasTenant.xmpl>location.locn|>birthPlace.madsrdf
```

Does this make sense?

docs/API.TraversingOutboundAndInboundLinksWithXgAndYg.md
nruppin: OK, I'm cool with the yg example, now could you please provide a good example for pipes in gqp?

docs/API.TraversingOutboundAndInboundLinksWithXgAndYg.md
```diff
@@ -42,7 +42,7 @@ You do this by using the **xg** and **yg** parameters described in the following
 
 The following query searches for Organization infotons whose name contains "Marriott Ownership" and whose city contains "Orlando":
 
-    <cm-well-host>/permid.org?op=search&qp=CommonName.mdaas:Marriott%20Ownership,organizationCity.mdaas:Orlando&with-data&format=ttl&length=1
+    <cm-well-host>/permid.org?op=search&qp=CommonName.mdaas:Marriott%20Ownership,organizationCity.mdaas:Orlapipndo&with-data&format=ttl&length=1
```
hochgi: umm... typo?

docs/API.TraversingOutboundAndInboundLinksWithXgAndYg.md
```diff
@@ -287,9 +287,9 @@ This query retrieves all infotons that point to the matched infotons through the
 
 If you want to expand along several paths using a single query, you can add several expansion expressions separated by the pipe ('|') operator.
 
-For example, the following query searches for organizations whose name contains "Marriott", and also retrieves Quote and Instrument infotons that are issues by those organizations. Note the use of the '|' operator to expand along two different inbound links.
+For example, in a single query the following call retrieves both the place that John Smith was born (via an outbound link), and all the places where John Smith lived (via a combination of inbound and outbound links):
 
-    <cm-well-host>/permid.org?op=search&qp=CommonName.mdaas:Marriott&yg=<isQuoteOf.mdaas|<isIssuedBy.mdaas
+    <cm-well-host>/example.org/Individuals/John_Smith?yg=<hasTenant.xmpl>location.locn|>birthPlace.madsrdf
```
hochgi: We probably want to reference a RDF sample here...
The query alone has no context. You can modify the sample I gisted, or use something of your own...
Doesn't really matter, but data sample as context is needed here IMO.

docs/API.ReturnCodes.md
```diff
@@ -9,5 +9,6 @@ Return Code | Description
 204 | No more results
 404 | Infoton not found
 401 | Unauthorized user, or password does not match user name
+422 | The request has no effect on CM-Well data
```
hochgi: a more accurate phrasing is: "Request is semantically correct, but does not make sense in CM-Well"

docs/API.TraversingOutboundAndInboundLinksWithXgAndYg.md
hochgi: e.g: for the president example, let's say I want to search all the US presidents that either were born outside the US, or were living in the white house before 2000.  
`gqp` would be:
```
gqp=<hasTenant.xmpl[hasBeginning.time<2000-01-01]>location.locn[name.geonames::The White House]|>birthPlace.madsrdf[-countryCode::US]
```

PR #305, opened 2017-11-20
JAVA_HOME env var

hochgi:


server/cmwell-cons/app/cmw.env
```diff
@@ -1,6 +1,7 @@
 export PATH=`pwd`/components-extras/scala/bin:$PATH
 export PATH=`pwd`/components-extras/java/bin:$PATH
 export PATH=`pwd`/bin/utils:$PATH
+export JAVA_HOME=`pwd`/components-extras/java
```
e-orz: cm-well doesn't always deploy itself java.
Changing this blindly will cause systems without deployed java to have a wrong JAVA_HOME

hochgi: You're absolutely right.
What do you suggest instead then...?

bryaakov: Just a thought, "getOrElseExport" will do the trick here?
Something like ```[ -z "$JAVA_HOME" ] && export JAVA_HOME=`pwd`/components-extras/java```

hochgi: Problem is with servers that has env var defined to old java, and it takes precedence over PATH.

e-orz: The simple solution will be to check whether java is deployed with cm-well and only then set the JAVE_HOME

PR #308, opened 2017-11-21
Enable sbt ws console

bryaakov:
_// I wonder why we had the `pack`'s one in the first place..._

### DIFF
--- a/server/cmwell-ws/build.sbt
+++ b/server/cmwell-ws/build.sbt
@@ -94,6 +94,4 @@ mappings in Universal += {
 
 fullTest := (test in Test).value
 
-fullClasspath in (Compile,console) := {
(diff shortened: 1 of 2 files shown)


server/cmwell-ws/build.sbt
```diff
@@ -95,5 +95,5 @@ mappings in Universal += {
 fullTest := (test in Test).value
 
 fullClasspath in (Compile,console) := {
-  (fullClasspath in (Compile,console)).value :+ Attributed.blank(packResourceDir.value.keys.head)
+  (fullClasspath in (Compile,console)).value :+ Attributed.blank(sourceDirectory.value / "cws" / "resources")
```
hochgi: could be simplified to:
```
fullClasspath in (Compile,console) += Attributed.blank(sourceDirectory.value / "cws" / "resources")
```

PR #315, opened 2017-11-27
explicitly adding HTTP scheme when no scheme provided, but port is

### LINKED ISSUE #309: SimpleHttpClient throws an exception
In case of a malformed request (without "http://") an exception is thrown instead of a failed future.

hochgi:
Which causes akka's UriParser to fail: https://github.com/akka/akka-http/issues/1547 
This commit fixes #309

### DIFF
--- a/server/cmwell-util/src/main/scala/cmwell/util/http/SimpleHttpClient.scala
+++ b/server/cmwell-util/src/main/scala/cmwell/util/http/SimpleHttpClient.scala
@@ -136,15 +136,36 @@ object SimpleHttpClient extends LazyLogging {
   }.toList
 
   private def mkURI(uri: String, queryParams: Seq[(String,String)]) = {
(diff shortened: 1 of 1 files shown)


server/cmwell-util/src/main/scala/cmwell/util/http/SimpleHttpClient.scala
```diff
@@ -136,15 +136,36 @@ object SimpleHttpClient extends LazyLogging {
   }.toList
 
   private def mkURI(uri: String, queryParams: Seq[(String,String)]) = {
-    if (queryParams.isEmpty) uri
+
+    val noSchemeWithPort = uri.matches("\\w+:\\d+")
```
e-orz: \\w does include digits but not hyphen or dot and those won't be caught properly

PR #319, opened 2017-11-30
Rewrite AuthCache as a Map

bryaakov:
* Load eagerly (with a certain delay + retries)
* When falling back to direct read - add retrieved key-value to map.

### DIFF
--- a/server/cmwell-ws/app/security/EagerAuthCache.scala
+++ b/server/cmwell-ws/app/security/EagerAuthCache.scala
@@ -1,0 +1,125 @@
+/**
+* Copyright 2015 Thomson Reuters
+*
+* Licensed under the Apache License, Version 2.0 (the “License”); you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*   http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+* an “AS IS” BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+*
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+package security
+
+import javax.inject.{Inject, Singleton}
+
+import com.typesafe.scalalogging.LazyLogging
+import logic.CRUDServiceFS
+import play.api.libs.json.{JsValue, Json}
+
+import scala.concurrent.duration.DurationInt
+import scala.concurrent.{Await, ExecutionContext, Future}
+import scala.util.Try
+import cmwell.domain.{Everything, FileContent, FileInfoton, Infoton}
+import cmwell.fts.{PaginationParams, PathFilter}
+import cmwell.util.concurrent._
+
+@Singleton
+class EagerAuthCache @Inject()(crudServiceFS: CRUDServiceFS)(implicit ec: ExecutionContext) extends LazyLogging {
+  private[this] var oData: AuthData = AuthData.empty
+  private[this] var nData: AuthData = AuthData.empty
+
+  private def data(nbg: Boolean) = if(nbg) nData else oData
+
+  SimpleScheduler.scheduleAtFixedRate(0.seconds, 10.minutes)(load())
+
+  // TODO Do not Await.result... These should return Future[Option[JsValue]]]
+  def getRole(roleName: String, nbg: Boolean): Option[JsValue] = {
+    data(nbg).roles.get(roleName).orElse {
+      Await.result(directReadFallback(s"/meta/auth/roles/$roleName", nbg), 6.seconds).map { role =>
+          logger.info(s"AuthCache(nbg=$nbg) role $roleName was not in memory, but added to Map")
+          if(nbg)
+            nData = nData.copy(roles = nData.roles + (roleName -> role))
+          else
+            oData = oData.copy(roles = oData.roles + (roleName -> role))
+        role
+      }
+    }
+  }
+
+  def getUserInfoton(userName: String, nbg: Boolean): Option[JsValue] = {
+    data(nbg).roles.get(userName).orElse {
+      Await.result(directReadFallback(s"/meta/auth/users/$userName", nbg), 6.seconds).map { user =>
+        logger.info(s"AuthCache(nbg=$nbg) user $userName was not in memory, but added to Map")
+        if(nbg)
+          nData = nData.copy(users = nData.users + (userName -> user))
+        else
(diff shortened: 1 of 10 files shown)


server/cmwell-ws/app/security/EagerAuthCache.scala
```diff
@@ -0,0 +1,125 @@
...
+        else
+          oData = oData.copy(users = oData.users + (userName -> user))
+        user
+      }
+    }
+  }
+
+  def invalidate(nbg: Boolean): Unit = load()
+
+  private def load(): Unit = {
```
hochgi: `load()` can be invoked from periodic scheduling or users calling `invalidate()`. It encapsulate a possibly long lasting effect. Caller has no way to cancel execution. Frequent calls to `invalidate()` may cause thread leak (sort of...).  

I suggest we guard the shared state updates. Maybe with something like an agent (old akka style), or similar construct. In the long run we can implement a comsume based reactive cache (some foundations can be found in `PassiveFieldTypesCache`, which we may want to generalize and reuse here)

server/cmwell-ws/app/security/EagerAuthCache.scala
```diff
@@ -0,0 +1,125 @@
...
+    retryUntil[AuthData](isSuccessful = !_.isEmpty, maxRetries = 30, delay = 10.seconds)(loadOnce(false)).map { data => oData = combiner(oData, data) }
```
hochgi: 10 retries should be enough IMO.
also, if ES is unresponsive, FTS' future is timed out after 10 sec. you schedule 10 sec after failure the next invokation.
This results in (10 + 10 + delta) seconds per retry.
multiply by 30, you'll get over 10 minutes.

server/cmwell-ws/app/security/EagerAuthCache.scala
```diff
@@ -0,0 +1,125 @@
...
+    delayedTask(5.seconds)(
+      retryUntil[AuthData](isSuccessful = !_.isEmpty, maxRetries = 30, delay = 10.seconds)(loadOnce(true)).map { data => nData = combiner(nData, data) }
```
hochgi: same here regarding the numbers... let's avoid overlapping retry invocations.

server/cmwell-ws/app/security/EagerAuthCache.scala
```diff
@@ -0,0 +1,125 @@
...
+import cmwell.util.concurrent._
+
+@Singleton
+class EagerAuthCache @Inject()(crudServiceFS: CRUDServiceFS)(implicit ec: ExecutionContext) extends LazyLogging {
+  private[this] var oData: AuthData = AuthData.empty
+  private[this] var nData: AuthData = AuthData.empty
+
+  private def data(nbg: Boolean) = if(nbg) nData else oData
+
+  SimpleScheduler.scheduleAtFixedRate(0.seconds, 10.minutes)(load())
```
hochgi: can we change to 15 min?

server/cmwell-ws/app/security/EagerAuthCache.scala
```diff
@@ -0,0 +1,125 @@
...
+        val (path, payload) = t
+        val isUser = path.startsWith("/meta/auth/users")
+        val key = path.substring(path.lastIndexOf("/") + 1)
+        if (isUser) Left(key -> payload)
+        else Right(key -> payload)
+      }
+      logger.info(s"AuthCache(nbg=$nbg) Loaded with ${usersData.size} users and ${rolesData.size} roles.")
+      AuthData(usersData.toMap, rolesData.toMap)
+    }.recover { case t: Throwable =>
+      logger.info(s"AuthCache(nbg=$nbg) failed to load because ${t.getMessage}")
```
hochgi: don't hide the error. let's change the log to:
`logger.error(s"AuthCache(nbg=$nbg) failed to load", t)`

server/cmwell-ws/app/security/EagerAuthCache.scala
```diff
@@ -64,13 +68,20 @@ class EagerAuthCache @Inject()(crudServiceFS: CRUDServiceFS)(implicit ec: Execut
...
-  def invalidate(nbg: Boolean): Unit = load()
+  def invalidate(nbg: Boolean): Future[Boolean] = if(isLoadingSemaphore) Future.successful(false) else load(nbg)
 
-  private def load(): Unit = {
-    retryUntil[AuthData](isSuccessful = !_.isEmpty, maxRetries = 30, delay = 10.seconds)(loadOnce(false)).map { data => oData = combiner(oData, data) }
-    delayedTask(5.seconds)(
-      retryUntil[AuthData](isSuccessful = !_.isEmpty, maxRetries = 30, delay = 10.seconds)(loadOnce(true)).map { data => nData = combiner(nData, data) }
-    )
+  private def load(nbg: Boolean): Future[Boolean] = {
+    isLoadingSemaphore = true
```
hochgi: where is it tested?
I'm expecting a: `if(isLoadingSemaphore) { /* do something */ }` somewhere...
Sill thinks `atomicBoolean.testAndSet` is better suited here, in case it's gonna be used that is.

server/cmwell-ws/app/security/EagerAuthCache.scala
```diff
@@ -33,9 +33,12 @@ class EagerAuthCache @Inject()(crudServiceFS: CRUDServiceFS)(implicit ec: Execut
   private[this] var oData: AuthData = AuthData.empty
   private[this] var nData: AuthData = AuthData.empty
 
+  private[this] var isLoadingSemaphore: Boolean = false
```
hochgi: isn't `AtomicBoolean` more suited here?

server/cmwell-ws/app/security/EagerAuthCache.scala
bryaakov: it is tested in the `invalidate` method.
Do we really need the AtomicBoolean here?

hochgi: why only protect against one side?
If you `invalidate()` a nanosec before scheduler is invoked, you'll get concurrent updates.
also, an abusive (malicious or buggy) user can `invalidate()` multiple times concurrently. you need to protect against that as well.

you named it "semaphore". either it is, or it's not. 
use an atomic boolean (shouldn't be called often anyway, overhead can't hurt),
or don't use anything at all.

at least that what I think. in any case, I'm okay with going with it as is, as long as we are aware it is not really protected, (inline docs?)

AnnaGrebnev:
@yanpom please perform the automation to check it

PR #323, opened 2017-12-03
Dc sync of current infotons only without history

e-orz:
1. add with-history parameter to the sync infoton (default is true)
2. add support for with-history to /proc/dc/someId (default is false)
3. changed the dc monitory to reflect the changes and handle with-history properly

### DIFF
--- a/server/cmwell-dc/src/main/scala/cmwell/dc/stream/DataCenterSyncManager.scala
+++ b/server/cmwell-dc/src/main/scala/cmwell/dc/stream/DataCenterSyncManager.scala
@@ -35,7 +35,9 @@ import scala.concurrent._
 import ExecutionContext.Implicits.global
 import scala.concurrent.duration._
 import scala.util.matching.Regex
+import scala.util.parsing.combinator.JavaTokenParsers
 import scala.util.{Failure, Success, Try}
+import scala.language.reflectiveCalls
 
 /**
   * Created by gilad on 1/4/16.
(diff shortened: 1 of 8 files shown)


server/cmwell-dc/src/main/scala/cmwell/dc/stream/DataCenterSyncManager.scala
```diff
@@ -352,14 +376,20 @@ class DataCenterSyncManager(dstServersVec: Vector[(String, Option[Int])], manual
...
+          }
           val fromIndexTime = f.asJsObject.fields.get("fromIndexTime").map {
             case JsArray(Vector(JsNumber(idxTime))) => idxTime.toLong
           }
           val tsvFile = f.asJsObject.fields.get("tsvFile").map {
             case JsArray(Vector(JsString(file))) => file
           }
-          val qpStr = userQp.fold("")(qp => s"?qp=$qp")
-          DcInfo(s"$dataCenterId$qpStr", location, fromIndexTime, tsvFile = tsvFile)
+          val qpStr = userQp.fold("")(qp => s"qp=$qp")
```
hochgi: very minor optimization:
`val qpStr = userQp.fold("")("qp=".+)`

server/cmwell-dc/src/main/scala/cmwell/dc/stream/DataCenterSyncManager.scala
```diff
@@ -352,14 +376,20 @@ class DataCenterSyncManager(dstServersVec: Vector[(String, Option[Int])], manual
...
+          val whStr = userWh.fold("with-history")(wh => if (wh == "true") "with-history" else "")
+          val qpAndWhStr = (for (str <- List(qpStr, whStr) if str.nonEmpty) yield str).mkString("&")
```
hochgi: why not simply `List(qpStr, whStr).filterNot(_.isEmpty)`

server/cmwell-fts/src/main/scala/cmwell/fts/FTSServiceES.scala
```diff
@@ -1036,11 +1036,13 @@ class FTSServiceES private(classPathConfigFile: String, waitForGreen: Boolean)
     }
   }
 
-  def getLastIndexTimeFor(dc: String, partition: String = defaultPartition, fieldFilters: Option[FieldFilter])
+  override def getLastIndexTimeFor(dc: String, withHistory: Boolean, partition: String = defaultPartition, fieldFilters: Option[FieldFilter])
                          (implicit executionContext: ExecutionContext) : Future[Option[Long]] = {
-
+    val partitionsToSearch =
+      if (withHistory) List(partition + "_current", partition + "_history")
```
hochgi: may be something like:
```scala
val request = {
  val req = if(withHistory) client.prepareSearch(partition + "_current", partition + "_history")
            else client.prepareSearch(partition + "_current")

  req.setTypes("infoclone")
     .addFields("system.indexTime")
     .setSize(1)
     .addSort("system.indexTime", SortOrder.DESC)
}
```

PR #338, opened 2017-12-10
exposing "parallelism" query parameter for sstream & bulk-consume

hochgi:
cc @Ivan-Shestakov

### DIFF
--- a/server/cmwell-ws/app/controllers/Application.scala
+++ b/server/cmwell-ws/app/controllers/Application.scala
@@ -699,6 +699,7 @@ callback=< [URL] >
         val withDeleted = request.queryString.keySet("with-deleted")
         val withMeta = request.queryString.keySet("with-meta")
         val length = request.getQueryString("length").flatMap(asLong)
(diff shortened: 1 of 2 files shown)


AnnaGrebnev:
@Ivan-Shestakov What was implemented:
The op=sstream API (which bulk-consume uses behind the scenes) was originally takes the advantage of each and every Elasticsearch index exists in the Cluster. So technically, the more data you have the faster it is. However, this has a downside – on a cluster with too many indices, this will blow up the system with tons of connections and won’t work at all. So we added a limitation to the amount of indices that are being streamed at a given moment. Our default configurable value is 10, but it can be changed per request. API: ?op=[sstream|bulk-consume]&parallelism=<INTEGER>

What to test:
1. To run sanity test
2. To run bulk-consume automation
3. To extend sstream test to check possible options of parallelism (by default it's 10): API: ?op=[sstream|bulk-consume]&parallelism=<INTEGER>

Ivan-Shestakov:
From initial testing of parallel sstreams I see that sstream without data is the most load producing and most sensitive (or breakable) of all stream apis. Reducing parallelism to 5 and even 3 - reduces but not eliminates failed stream percentages.

PR #339, opened 2017-12-11
Dc-Sync: ignore infotons with a faulty index time in CAS

e-orz:
1. Remove the error printed when TSV retrieve failed. It's only a warning and can be retried.
2. Check that the index time from ES and from CAS match. If not ignore this infoton and write it to the red log.
3. Add the relevant with-history field for the dc virtual infotons in /proc/dc.
4. remove the full stack trace in log print for missing uuid exception


AnnaGrebnev:
@yanpom please talk to me how to simulate the "broken" data

e-orz:
@AnnaGrebnev It will be time consuming to check it and it's already running (checked when we synced production to the new one). I think it can be closed as checked.
One thing to note. There were 80% errors there and the sync was very slow.

AnnaGrebnev:
@e-orz done, I've closed it.

PR #342, opened 2017-12-11
a fix for stream engine (backing sstream & bulk-consume)

hochgi:


server/cmwell-ws/app/cmwell/ws/Streams.scala
```diff
@@ -205,6 +205,7 @@ class Streams @Inject()(crudServiceFS: CRUDServiceFS) extends LazyLogging {
 
   private def singleScrollSourceHandler(nbg: Boolean, withDeleted: Boolean, ec: ExecutionContext)
                                        (firstHit: IterationResults): Source[IterationResults,NotUsed] = Source.unfoldAsync(firstHit) {
+    case ir@`firstHit` => crudServiceFS.scroll(ir.iteratorId, 60, withData = false, nbg || withDeleted).map(iir => Some(iir -> ir))(ec)
```
e-orz: if the first hit has no result, won't it end in an endless loop?

hochgi: no... iteratorId is different

PR #345, opened 2017-12-13
Moving some log.info calls to log.debug, to reduce noise in logs

bryaakov:


server/cmwell-ws/app/controllers/Application.scala
```diff
@@ -855,7 +855,7 @@ callback=< [URL] >
                             if (bytes.isEmpty) ""
                             else all.lines.next()
                           }
-                          logger.info(s"""[$id] onPush(first line: "$elem", num of lines: ${all.lines.size}, num of chars: ${all.length})""")
+                          logger.debug(s"""[$id] onPush(first line: "$elem", num of lines: ${all.lines.size}, num of chars: ${all.length})""")
```
hochgi: only if `debugLog==true`, and we want it info level

bryaakov: didn't realize it was only when debugLog==true. will fix it.

PR #352, opened 2017-12-20
"ironing" dots to underscore

hochgi:
converting dots to underscore to safe guard from ES unintended nesting.
change was checked locally using:
```
$ curl localhost:9000/_in?format=ntriples -H "X-CM-WELL-TOKEN: ... " --data-binary '<http://some.made/up/domain> <http://and.made.up/field#with.dots> "and value" .'
{"success":true}
$ curl -s localhost:9000/proc/fields?format=json | jq -r '.fields.fields[]' | grep -E 'with(_|.)dots'
with_dots.eadC6A:string
$ curl -s 'localhost:9000/?op=search&recursive&qp=with.dots.$eadC6A:&format=json' | jq '[.results.total,.results.infotons[0].system.path]'
[
  1,
  "/some.made/up/domain"
]
$ curl -s 'localhost:9000/?op=search&recursive&qp=with_dots.$eadC6A:&format=json' | jq '[.results.total,.results.infotons[0].system.path]'
[
  1,
  "/some.made/up/domain"
]
```

### DIFF
--- a/server/cmwell-common/src/main/scala/cmwell/common/formats/AbstractJsonSerializer.scala
+++ b/server/cmwell-common/src/main/scala/cmwell/common/formats/AbstractJsonSerializer.scala
@@ -40,14 +40,14 @@ class AbstractJsonSerializer {
 trait NsSplitter {
   def splitNamespaceField(field: String) = field.lastIndexOf(".") match {
     case -1 => "nn" -> field
-    case i => field.substring(i+1) -> field.substring(0,i)
+    case i => field.substring(i+1) -> field.substring(0,i).replace('.','_')
   }
 
   def reverseNsTypedField(field: String) = {
     if(field.startsWith("system.") || field.startsWith("content.") || field.startsWith("link.")) field
     else {
       val (ns, typedField) = splitNamespaceField(field)
-      "fields." + ns + "." + typedField
+      "fields." + ns + "." + typedField.replace('.','_')
     }
   }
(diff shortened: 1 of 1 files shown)


AnnaGrebnev:
@GuyGuterman please add list of test cases and implement it in the sanity automation

GuyGuterman:
1. To check with ingesting (NTRIPLES format) when at predicate have a dot, for example, "hello.hello". The result can be seen in ES dots will be changed to underscore.
2. Repeat step 1 with, for example, "hello_hello". The result will be as step 1.
3. Repeat step 1 with NQUADS. The result will be as step 1.
4. Repeat step 2 with NQUADS format, example, "hello_hello". The result can as step 2.
5. Repeat step 1 when at predicate have a symbol/s, for example, h$ello_h$ello. The result will be "predicate namespace must not contain a dollar".
6. Repeat step 5 NQUADS format when at predicate have a symbol/s, for example, h$ello_h$ello. The result will be as step 5.
6. Repeat step 1 when at predicate have more than one dot, for example, he.llo.he.llo. The result will be as at step 1.
7. Repeat step 6 when at predicate have more than one underscore, for example, he_llo_he_llo. The result can be seen in ES dots will be changed to underscore.
8. Repeat step 1 when at predicate will be more then dots and underscore together, for example, he.llo_he.llo. The result can be seen in ES dots will be changed to underscore.

PR #358, opened 2017-12-25
forbiding writes (regular non _ow writes) of deleted infotons

hochgi:


GuyGuterman:
1. To check ingest infoton check if is received. Then Delete him and try to ingest DeletedInfoton the result will be "Writing a DeletedInfoton does not make sense. use proper delete API instead" and will not be accepted.
2. To check ingest few infotons when one of them DeletedInfoton. the result will be the same as step 1.

AnnaGrebnev:
@hochgi what's describe deleted infoton? the same UUID?

hochgi:
> @hochgi what's describe deleted infoton? the same UUID?
No. this is a deleted infoton:
```
<http://some.path.in/cm/well>
        sys:dataCenter    "dcid" ;
        sys:indexTime     "1234567891011"^^xsd:long ;
        sys:lastModified  "2000-01-01T00:00:00.000Z"^^xsd:dateTime ;
        sys:parent        "/some.path.in/cm" ;
        sys:path          "/some.path.in/cm/well" ;
        sys:type          "DeletedInfoton" ;
        sys:uuid          "12341234123412341234123412341234" .
```
what matters is: `sys:type "DeletedInfoton"` property.

Issue #328, opened 2017-12-06
iterate API should ease exposing internal ES scroll id


hochgi:
`handleScroll` should take a `debug-info` param, and add the original scroll id in payload.
Also, we should add an API for `try-to-continue-iteration` with ES scroll ids supplied as payload. 
It will enable a user to maybe recover a failed iteration. Lastly every failed response for iterate API should return original ES scroll id in payload to enable that by default.

AnnaGrebnev:
@yanpom please prepare list of test cases and extend the sanity automation

AnnaGrebnev:
What to test:
1. Make sure no regression in functionality of "create-iterator" and "next-chunk" API, 
2. Use that API via HAProxy when ES is stressed.

hochgi:
@AnnaGrebnev  this is not yet implemented

GuyGuterman:
@hochgi Any updates about this issue? will be implemented in Quetzal?

hochgi:
No... it's low priority.

PR #363, opened 2017-12-26
Keeping iterator actor until it is safe to dispose it

bryaakov:
Related: Issue #328


GuyGuterman:
Validated that on the creation of new iterator, responded with a token. Then stopped ES. Tried to "use" this token and no response (as expected). Rerun ES and used the SAME token again. Returned as expected. Repeated few times.
Moving to Done

PR #377, opened 2018-01-01
Update docs according to #363

bryaakov:
Related PR: #363

### DIFF
--- a/docs/API.Stream.CreateIterator.md
+++ b/docs/API.Stream.CreateIterator.md
@@ -49,7 +49,6 @@ session-ttl | The time, in seconds, until the iteration session expires. The ite
 
 ## Notes ##
 
-* If the iteration process fails in the middle for any reason, you will have to restart the process from the beginning (that is, iterate again over all infotons that match the query).
 * An alternative is to use the **consumer** API (see **Related Topics**), which allows you to save the iteration state and restart from the same point after a failure.
 
 ## Related Topics ##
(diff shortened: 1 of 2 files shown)


docs/API.Stream.CreateIterator.md
```diff
@@ -49,7 +49,6 @@ session-ttl | The time, in seconds, until the iteration session expires. The ite
 
 ## Notes ##
 
-* If the iteration process fails in the middle for any reason, you will have to restart the process from the beginning (that is, iterate again over all infotons that match the query).
```
hochgi: If the iteration process fails in the middle for any reason, **you might be able to resume with last token within the TTL window, otherwise you'll have to** restart the process from the beginning (that is, iterate again over all infotons that match the query).

docs/API.Stream.GetNextChunk.md
```diff
@@ -94,7 +94,7 @@ format | The output format of the streamed chunks. The default value is **ntripl
 
 ## Notes ##
 
-* **If the iteration process fails in the middle for any reason, you will have to restart the process from the beginning (that is, iterate again over all infotons that match the query).**
+* **If the iteration process fails in the middle for any reason, you can retry the same request, as long as `session-ttl` has not passed.**
```
hochgi: and in case the last token is no longer valid, you will have to restart the process from the beginning (that is, iterate again over all infotons that match the query).

docs/API.Stream.CreateIterator.md
nruppin: Not sure I understand this. What is the "last token within the TTL window"? Can't we simply say the last token we received? Also, what does "you might be able" mean? I thought the purpose of the feature was to enable this. What does failure/success depend on? Are we saying that we recommend doing a retry, and if it fails, start from the beginning?

PR #397, opened 2018-01-16
BufferFillerActor now always decodes Gzip responses regardless of the…

nicklester:
BufferFillerActor now always decodes Gzip responses regardless of the Http status code. Decompression logic moved to new utility class


server/cmwell-data-tools/src/main/scala/cmwell/tools/data/downloader/consumer/Downloader.scala
```diff
@@ -405,12 +405,15 @@ class Downloader(baseUrl: String,
...
+      val gzipContentEncoding = `Content-Encoding`(HttpEncodings.gzip)
+
       HttpRequest (
         uri = s"${formatHost(baseUrl)}/_out?format=$format$paramsValue",
         method = HttpMethods.POST,
         entity = HttpEntity(concatByteStrings(paths, ByteString(",\n")).utf8String)
-          .withContentType(ContentTypes.`text/plain(UTF-8)`)
-      ).addHeader(RawHeader("Accept-Encoding", "gzip"))
+          .withContentType(ContentTypes.`text/plain(UTF-8)`),
+        headers = scala.collection.immutable.Seq(gzipContentEncoding)
```
e-orz: This is the wrong header.
1.  This header will tell the server that the request is gzipped (which is not, I think) and the server won't parse it properly.
1. The server won't know that it can gzip the response and it won't be gzipped.

PR #414, opened 2018-01-23
converted new FTS to no timeouts by default

hochgi:


Ivan-Shestakov:
Performance baseline passed. Moving to Done.

PR #419, opened 2018-01-24
Dashboards batch renaming

hochgi:


e-orz:
@bryaakov should we close this branch?

Issue #421, opened 2018-01-24
imp crashes due to unexpected timeouts from cassandra


hochgi:
This should be fixed in 2 stages:

- [ ] first, as a hotfix, IRW service should take a longer timeout for bg. unlike ws, it should not be considered interactive, and large infotons (e.g. accumulated merges) is OK to be served longer. instead of the 10 seconds defined, we can safely enlarge to 2 minutes timeout (only in bg)
- [ ] second phase, we should do a deeper refactoring effort, and try to expose cassandra operations reactively, in a backpressured fashion. instead of a full read, we can do a smarter streamed read.

The second phase relates to #420

PR #422, opened 2018-01-24
enlarging cas timeouts to 120 seconds in bg 

hochgi:
This should resolve the first phase of #421


Ivan-Shestakov:
After discussing with @hochgi - no simple way to simulate this. Covered by our performance tests.
Moving the issue to Done.

PR #432, opened 2018-01-25
change qstream default size to 3000

e-orz:


GuyGuterman:
Tested:

Search param | Expected result | Actual result
-- | -- | --
Tested with used curl -v --raw and default length-hint (not registered as parameter)  | about count 3000 | Result was ~127
Tested with used curl -v --raw and default length-hint equal to 50 | about count 50  | Result was ~37
Tested with used curl -v --raw and default length-hint equal to 1500 | about count 1500 | Result was ~127
Tested with used curl -v --raw and default length-hint equal to 3000 | about count 3000 | Result was ~127

> In the button line the test failed and therefore a bug was opened accordingly

[New bug #485](https://github.com/thomsonreuters/CM-Well/issues/485)

GuyGuterman:
After discussing with @hochgi - no simple way to simulate this. Changed between ES and WS in the code.
Moving the issue to Done.

PR #440, opened 2018-01-28
adding configuration files for cons (application, logback)

e-orz:
This will solve the issue of the logs printing to stdout when using the cons.


Ivan-Shestakov:
Validated that the following files were created in /home/user/cmwell-install-package folder (from which cons is launched). 
akka-log-2017-06-25.0.gz  
akka.log  
alerts.log  
application-log-2017-06-25.0.gz  
application.log  
cluster_state.log  
red.log  
yellow.log
Moving to Done.

PR #452, opened 2018-01-30
remove the type check in DcChecker

e-orz:
In case there was no sync in the sytem the type field won't exit.
It will solve the field not found exception in the logs every 30 seconds


GuyGuterman:
Action | Expected | Actual
-- | -- | --
With qp=type::remote | Start working  | as expected
Without qp=type::remote | Start working | as expected

Moving to Done.

Issue #451, opened 2018-01-29
STP Dashboard - Point in time stays 1970-01-01T00:00	


yanpom:
Sensor is working,pulling and materilizing infotons but its point in time remains 
1970-01-01T00:00

Also, its not quit understood what point of time is  - we should propably to rename it to remote index time (as in dcsync) or some more propriate name

yanpom:
Please see attached print screen

nicklester:
Token is saved after reading so many paths, or up until a certain time window has elapsed. So until  a token has been saved I think the token will be read as 0, which interprets to 1970-01-01T00:00 when converted to a date. 

What should we do in this circumstance? Show empty field? Or some text? @yanpom  / @YoniMataraso

e-orz:
The token is actually the last persisted time (the time STP will start the sensor from). A time of 0 is actually saying it will start from the beginning which is the truth...

nicklester:
Correct. Just wondering if we should show an empty field rather than 1970.... I will change this and submit PR - and we can change logic if need be :)

Issue #446, opened 2018-01-29
STP Dashboard - Status Enhancements


nicklester:
From merging #370 some other suggestions proposed by @e-orz 

* JobPausing/JobStopping statuses should also show in the dashboard. 
* Any exceptions thrown by a failed job should show in the dashboard

PR #454, opened 2018-01-30
Dashboard feedback fixes

nicklester:
Addressing feedback from @e-orz in issue #446. Job exceptions will be displayed and the status of Pausing/Stopping jobs is also displayed.

Addressed feedback from @yanpom in issue #451 and tidied up tables


server/cmwell-sparql-agent/src/main/scala/cmwell/tools/data/sparql/SparqlProcessorManager.scala
```diff
@@ -64,27 +64,26 @@ case object CheckConfig
...
-  override def statusString = " Active"
   val reporter: ActorRef
 }
 
-case class JobRunning(job: Job, killSwitch: KillSwitch, reporter: ActorRef) extends JobActive
-case class JobPausing(job: Job, killSwitch: KillSwitch, reporter: ActorRef) extends JobActive
-case class JobStopping(job: Job, killSwitch: KillSwitch, reporter: ActorRef) extends JobActive
+case class JobRunning(job: Job, killSwitch: KillSwitch, reporter: ActorRef, statusString : String = "Running") extends JobActive
+case class JobPausing(job: Job, killSwitch: KillSwitch, reporter: ActorRef, statusString : String = "Pausing") extends JobActive
+case class JobStopping(job: Job, killSwitch: KillSwitch, reporter: ActorRef, statusString : String = "Stopping") extends JobActive
```
hochgi: @nicklester Does `statusString` can take any other value than the given default?
If not, and it seems that way, it would be better to define the class as, e.g:
```scala
case class JobRunning(…) extends JobActive {
  override val statusString: String = "Running"
}
```
If in other places in code you want to pattern match and get the extracted string, please use extractors. e.g:
```scala
object JobActive {
  def unapply(job: JobActive): Option[(job: Job, killSwitch: KillSwitch, reporter: ActorRef, statusString: String)] = 
    Some((job.job,job.killSwitch,job.reporter,job.statusString))
}
…
job match {
  case JobActive(job, killSwitch, reporter, "Running") => …
  case JobActive(job, killSwitch, reporter, pausingOrStopping) => …
  …
}
```
Also, better yet, if strings are constants. let's use a sealed trait + case objects instead.

P.S.
If you use extractor, maybe you won't need `case class`es. 
You could use regular (`final`) classes instead.
Also, as a good practice, every `case class` should also be marked `final`

Issue #392, opened 2018-01-10
STP Dashboard Enhancements


nicklester:
After a discussion with @AnnaGrebnev , @yanpom , @e-orz a list of initial enhancements for the STP dashboard has been determined:

* Each job has several statuses. The dashboard should show the status of each job
* Each job also has several sensors. The dashboard should show the status of each sensor
* Failure reason / exceptions should be displayed by the sensor to save the end user from having to navigate to the logs
* We should show last index time for each sensor
* Show number of processed / updated infotons
* Show current position of STP

e-orz:
my additions:
* Show number of processed / updated infotons - for each job separately (each job should unite all the sensor so it cannot be done for each sensor)
* Show current position of STP - has no meaning (I think) the position should be for each sensor

e-orz:
my additions:
* Show number of processed / updated infotons - for each job separately (each job should unite all the sensor so it cannot be done for each sensor)
* Show current position of STP - has no meaning (I think) the position should be for each sensor

nicklester:
Also add the infoton moving average rate removed as part of #450

nicklester:
- [ ] All Job Statuses reflected on the dashboard
- [ ] Show status of each sensor
- [ ] Exception text string now displays for a failed job
- [ ] Last index time shows for each sensor
- [ ] Infotons received show per agent
- [ ] Infotons ingested and failed (processed) show in dashboard
- [ ] Re-add the infoton moving average rate for infotons received

nicklester:
@AnnaGrebnev , @yanpom - I believe that a lot of the requirements for this ticket are now implemented. I created a checklist to track. Can you check?

PR #472, opened 2018-02-01
450

nicklester:
Hi @e-orz  - not sure if we want to do this, but Yan suggested this is the best metric.


nicklester:
Hi Yaakov - do you have time to review this?

e-orz:
What I don't like about it is the fact that in case there are no new updates the rate will drop to 0 (give the feeling of a stuck system) even though the system is still running ok.
It's the same issue with current one minute rate...

Maybe you can add both metrics... I really don't know.

nicklester:
I don't disagree, but I know @yanpom was very keen for it to work this way.

nicklester:
Maybe as part of the dashboard improvements  (#392) we could think about adding the old metric in again so that we track both? Don't know what else to do really given @yanpom  is not around at the moment

nicklester:
I'll put the old metric back for next release as discussed. Will add to the dashboard enhancements ticket #392

Issue #442, opened 2018-01-28
STP Engine - Source configuration change requires a restart/upgrade


yanpom:
In order to change physical view source name from local to prod or vice versa, there is a need to manually change reference.conf file which is located in cmwell-sparql-triggered-processor-agent_2.12-1.5.93.jar
Than in order to those changes apply, there is a need in upgrade of the system
This configuration should be less complicated and be a part of agent creation (maybe its should be one of the agent parameters)

nicklester:
Added a new configuration option in the yml:

hostUpdatesSource: <http url of source cmwell>

If this is set it will override the value set in reference.conf, otherwise it will use the value set in reference.conf

nicklester:
P.S - for changes to take effect you need to re-upload the yml and pause and un-pause the job. This will cause the config to be reloaded.

e-orz:
@nicklester Are you sure pause-resume will be enough? As far as I remember full removal of the STP yaml infoton and re-ingest of it is needed.

Anyway, Naomi should document it cc @nruppin

PR #480, opened 2018-02-05
#442 Added optional source host configuration for agent STP config 

nicklester:


nicklester:
Note - we need to update any documentation to reflect the new optional parameter in the yml agent configuration files

Issue #483, opened 2018-02-05
Span Kafka LogDirs over more than one node


bryaakov:
Also - reduce retention from one week to one day (24hrs)

Ivan-Shestakov:
Validated on 2 lab environments: one fresh install, and another upgrade. 
The manual change required was to update environment configuration to take a `List("/kafka//folder")` instead of `"/kafka/folder"` 
Sanity passed (except for known failures (424 errors)).

PR #484, opened 2018-02-05
Implement #483 - Allowing more than one Kafka Data dir

bryaakov:
Also - Reduce retention to 24 hours rather than a week. _// this was another requirement._

P.S. @YoniMataraso Thanks for you help with getting started with implementation!


Ivan-Shestakov:
@bryaakov lets chat about how to test that.

PR #512, opened 2018-02-18
fixed rare (first time and early) nsID conversions that fails 

hochgi:
because we don't wait for Future + remove jmx from datastax client + upgraded metrics lib


server/cmwell-ws/app/ld/cmw/CMWellRDFHelper.scala
```diff
@@ -92,14 +92,29 @@ class CMWellRDFHelper @Inject()(val crudServiceFS: CRUDServiceFS, injectedExecut
...
-    newestGreatestMetaNsCacheImpl.get(nsID,timeContext).value.flatMap {
-//      case None => throw ServerComponentNotAvailableException(s"Internal old API (hashToUrl) used on id [$nsID], which was not yet in cache. subsequent requests should succeed eventually. Call should be migrated to new API")
-      case /*Some(*/Success((url,_))/*)*/ => Some(url)
-      case /*Some(*/Failure(_: NoSuchElementException)/*)*/ => None
-      case /*Some(*/Failure(e)/*)*/ => throw e
-    }
+  def hashToUrl(nsID: String, timeContext: Option[Long]): Option[String] = {
+    val f = hashToUrlAsync(nsID, timeContext)
+    f.value.fold[Option[String]] {
+      if (nsID.matches(looksLikeHashedNsIDRegex))
```
e-orz: regex should be compiled in advance to get performance.
This check is run lots of times and should be optimized.

server/cmwell-ws/app/ld/cmw/CMWellRDFHelper.scala
```diff
@@ -135,13 +150,16 @@ class CMWellRDFHelper @Inject()(val crudServiceFS: CRUDServiceFS, injectedExecut
...
-    newestGreatestMetaNsCacheImpl.get(nsID, timeContext).value.flatMap {
-//      case None => throw ServerComponentNotAvailableException(s"Internal old API (hashToUrlAndPrefix) used on id [$nsID], which was not yet in cache. subsequent requests should succeed eventually. Call should be migrated to new API")
-      case /*Some(*/Success(urlAndPrefix)/*)*/ => Some(urlAndPrefix)
-      case /*Some(*/Failure(_: NoSuchElementException)/*)*/ => None
-      case /*Some(*/Failure(e)/*)*/ => throw e
-    }
+  def hashToUrlAndPrefix(nsID: String, timeContext: Option[Long]): Option[(String,String)] = {
+    val f = newestGreatestMetaNsCacheImpl.get(nsID, timeContext)
+    f.value.fold[Option[(String, String)]] {
+      if (nsID.matches(looksLikeHashedNsIDRegex))
```
e-orz: same as above, optimize the regex.

server/cmwell-ws/app/ld/cmw/CMWellRDFHelper.scala
hochgi: not a lot of times. headline does say "rare"...
to be more specific:
This is called only from formatting code, where you have an infoton retrieved from cassandra already,
and you need to convert internal field names with hash (ns IDs) to proper representation.
Now, each nsID should be fetched eagerly, and once it is fetched, `f.value` will always be a `Some` since the value is stored in a map, and you just wrap it with `Future,successful`.
So for a real infoton, you could possibly get here only in the rare case of ingesting an infoton with a new namespace, which is not in cache, and reading it fast enough before the consume logic fetches it for you. This happened in tests but IRL, I don't see it happening often. 
The other scenario is when you have a virtual infoton, like those we make in `/proc/*`. If the field name contain a `.`(dot character), The formatter tries to resolve it. And this is the reason why I added the check. we have "field names" the look like `<module>@<host>`, e.g: `bg@127.0.0.1` - so in this case the `.1` is then being tried to be resolved as identifier. of course it won't succeed, but at least this check saves us the 9 seconds wait. 

Bottom line, I agree we should strive for performance. but this case is pretty rare...

PR #517, opened 2018-02-19
remove context check in stream

e-orz:
It is a heavy request and that didn't have any benefit (the limit was too high anyway)


Ivan-Shestakov:
Discussed with @e-orz , we can't recreate this condition on our systems. 
Verified that no functional regression introduced.
Moving to done.

Issue #458, opened 2018-01-31
STP Dashboard - There is a need in another state of agent execution


yanpom:
Apart from Active and Non-Active, there is a need for another state in stp dashboard per agent/sensor
Incremental Updates/Horizon state is needed to show the user that agent finished its normal work and now is in waiting/per demand condition. In this case the infoton rate is not needed also (or we can use this field for incremental/waiting status.

nicklester:
We should add a new column to the agent table to show whether we are in Horizon

nicklester:
![screen shot 2018-02-20 at 13 57 31](https://user-images.githubusercontent.com/1490089/36427819-16c91222-1646-11e8-9281-1926f12cd927.png)

PR #520, opened 2018-02-20
#458 - Show Sensor horizon status on dashboard

nicklester:


server/cmwell-data-tools/src/main/scala/cmwell/tools/data/downloader/consumer/BufferFillerActor.scala
```diff
@@ -142,13 +144,20 @@ class BufferFillerActor(threshold: Int,
       tsvCounter += 1
 
     case GetData if buf.nonEmpty =>
-      sender ! buf.dequeue()
+     sender ! ( buf.dequeue() match {
```
e-orz: deque will never return None (it will throw NoSuchElementException that is the reason for the below `if buf.nonEmpty` check.

nicklester: It is only there as the Downloader.scala has a check for None after it receives the message from GetData... seems to be None is used to signal end of the stream... Maybe it is redundant?

e-orz: Well it's my mistake. The dequeue method will never return null but the element itself can be.  I looked in the code and found the following:
https://github.com/thomsonreuters/CM-Well/blob/f7da7b700e619f42c236f201d281cf100285a933/server/cmwell-data-tools/src/main/scala/cmwell/tools/data/downloader/consumer/BufferFillerActor.scala#L229
which will cause the element to be None for the Downloader.
This "signal" is more "smart" than the horizon because it is set only when the buffer is empty (e.g. the down stream already took all the elements). It can be used to notify of horizon+all TSVs were taken by the down stream.
I never knew about it, it can maybe improve the horizon feature but I will not insist on it.

server/cmwell-data-tools/src/main/scala/cmwell/tools/data/downloader/consumer/BufferFillerActor.scala
```diff
@@ -142,13 +144,20 @@ class BufferFillerActor(threshold: Int,
...
+        case None => None
+        case Some(tokenAndData) => Some(tokenAndData._1,tokenAndData._2,horizon)
+      } )
```
e-orz: BTW, it will be nicer to write:
`buf.dequeue().map(tokenAndData => Some(tokenAndData._1,tokenAndData._2,horizon))` which will result the same

server/cmwell-data-tools/src/main/scala/cmwell/tools/data/utils/akka/stats/DownloaderStats.scala
```diff
@@ -148,6 +153,12 @@ class DownloaderStats[T](isStderr: Boolean,
         }
       }
 
+      def setSensorHorizon(contextOption : Option[SensorContext]) = {
+        contextOption.map(context=>{
```
e-orz: for side effects the convention is to use `forEach` and not `map`

Issue #443, opened 2018-01-28
STP Agent - Physical view source machine name should be written in logs/dashboard


yanpom:
Following is written in agents.log on agent start:
018-01-28 14:34:35,533 -well-egl1-akka.actor.default-dispatcher-20] INFO  c.t.d.sparql.SparqlProcessorManager  - Got start request for job [job: industry, config name:None, sensors: industryindustry] from the user
2018-01-28 14:34:37,206 -well-egl1-akka.actor.default-dispatcher-20] INFO  c.t.d.sparql.SparqlProcessorManager  - starting job [job: industry, config name:None, sensors: industryindustry]
2018-01-28 14:34:37,228 -well-egl1-akka.actor.default-dispatcher-20] INFO  BufferFillerActor [industryindustry] - starting BufferFillerActor

There should be a physical view source machine name mentioned in the logs and also in dashboard. User should suppose to know from what environment data is pulled

nicklester:
Added source machine for STP to dashboard and to relevant STP logs. See screenshots

nicklester:
![screen shot 2018-02-14 at 20 05 51](https://user-images.githubusercontent.com/1490089/36225417-82fe2154-11c2-11e8-92d1-362b1e8dad12.png)
![screen shot 2018-02-14 at 20 05 26](https://user-images.githubusercontent.com/1490089/36225418-8320209c-11c2-11e8-8638-d44f894008aa.png)

PR #528, opened 2018-02-21
#443 Added STP source to dashboard and relevant logs

nicklester:


server/cmwell-dc/src/pack/logback.xml
```diff
@@ -158,7 +158,7 @@
             <cleanHistoryOnStart>true</cleanHistoryOnStart>
         </rollingPolicy>
         <encoder>
-            <pattern>%date{ISO8601} %-44.44([%thread]) %-5level %-36.36logger{36} - %msg%n</pattern>
+            <pattern>%date{ISO8601} %-44.44([%thread]) %-5level %-72.72logger{72} - %msg%n</pattern>
```
e-orz: changing the logger (module) length will break the log parser. see:
https://github.com/thomsonreuters/CM-Well/blob/f7da7b700e619f42c236f201d281cf100285a933/server/cmwell-cons/app/bin/utils/cm-well-formats.json#L9

The logger part of the was was already abused too much. It should be part of the message.

server/cmwell-data-tools/src/main/scala/cmwell/tools/data/utils/logging/DataToolsLogging.scala
```diff
@@ -26,9 +26,11 @@ trait DataToolsLogging {
   private[data] lazy val badDataLogger = LoggerFactory.getLogger("bad-data")
 
   val label: Option[String] = None
+  val baseUrl: String = ""
+
 
   protected lazy val logger: Logger = {
-    val loggerName = if (label.isEmpty) getClass.getName else s"${getClass.getName} [${label.get}]"
+    val loggerName = if (label.isEmpty) getClass.getName else s"${getClass.getName} [${label.get} ${baseUrl}]"
```
e-orz: According to Logback convention the logger is the module name.
I know it's not the case in STP but it shouldn't be worse than it is now, even better the current implementation should be changed.

PR #530, opened 2018-02-22
using the unmangled field name in ws in-memory filtering (xg/yg/gqp)

### LINKED ISSUE #529: [Regression] there is no full response when complicated YG query sends
1. Send complicated YG query where it is required to bring data from the different graphs (the example of real use case will be provided by email)
Expected results: full response from all graphs appear
Actual results: partial response appears (there is no on part of the graph)

Some troubleshooting steps that I did;
1. If I remove one predicate from the query, it works as expected 
2. the predicate is Boolean vs the rest
3. Search by this predicate works as expected for both cases: full path and prefix

It impacts negatively our customer.

hochgi:
and tests to cover it.
fixes #529


Ivan-Shestakov:
Verified by execution of complete automated regression suite + customer use cases.

PR #563, opened 2018-03-06
dc-sync now checks that the TSVs actually got match the number in the…

e-orz:
… X-CM-WELL-N header

### DIFF
--- a/server/cmwell-dc/src/main/scala/cmwell/dc/stream/TsvRetriever.scala
+++ b/server/cmwell-dc/src/main/scala/cmwell/dc/stream/TsvRetriever.scala
@@ -186,9 +186,9 @@ object TsvRetriever extends LazyLogging {
                 val expectedTsvCount = nHeader.value().toInt
                 val gotTsvCount = sortedData.size
                 if (gotTsvCount != expectedTsvCount) {
-                  val e = new Exception(s"Got $gotTsvCount but the expected count (taken from X-CM-WELL-N header) is $expectedTsvCount.")
+                  val e = new Exception(s"Got $gotTsvCount TSVs but the expected count (taken from X-CM-WELL-N header) is $expectedTsvCount.")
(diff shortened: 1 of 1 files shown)


server/cmwell-dc/src/main/scala/cmwell/dc/stream/TsvRetriever.scala
```diff
@@ -179,7 +179,21 @@ object TsvRetriever extends LazyLogging {
                 val consumeCount = Settings.initialTsvRetryCount - state.retriesLeft + 1
                 yellowlog.info(s"TSV (bulk)consume succeeded only after $consumeCount (bulk)consumes. token: ${state.tsvRetrieveInput}.")
               }
-              Success(TsvFlowOutput(sortedData, nextPositionKey, s.intValue == 204)) -> state.copy(lastException = None)
+              //for bulk-consume, X-CM-WELL-N header exists. Check that the number of TSVs got is the same as the number in the header.
+              h.find(_.name == "X-CM-WELL-N").fold{
+                Try(TsvFlowOutput(sortedData, nextPositionKey, s.intValue == 204)) -> state.copy(lastException = None)
+              } { nHeader =>
+                val expectedTsvCount = nHeader.value().toInt
```
hochgi: can fail. IMO better to do: `{ nHeader => Try{ … }.flatMap(identity)}`

server/cmwell-dc/src/main/scala/cmwell/dc/stream/TsvRetriever.scala
```diff
@@ -179,7 +179,21 @@ object TsvRetriever extends LazyLogging {
...
+                val gotTsvCount = sortedData.size
+                if (gotTsvCount != expectedTsvCount) {
+                  val e = new Exception(s"Got $gotTsvCount but the expected count (taken from X-CM-WELL-N header) is $expectedTsvCount.")
```
hochgi: you don't need `Exception` with stack trace here, since you use it as cause in a line below... 
it'll probably be better to use `NoStackTrace` trait instead

not very important though...

PR #580, opened 2018-03-14
fix the iterator scheduler to be at most 7 seconds and not 48 years

e-orz:


server/cmwell-ws/app/controllers/Application.scala
```diff
@@ -1502,7 +1502,8 @@ callback=< [URL] >
                 }
               }
 
-              val initialGraceTime = deadLine.timeLeft
+              //initialGraceTime can become negative (think of a lengthy GC that will cause now to be after the dead line), but it's ok because the scheduler will start immidiately
+              val initialGraceTime = Duration(deadLine - System.currentTimeMillis(), MILLISECONDS)
```
hochgi: is negative duration handled ok?
Even if so, do we want to use `Duration.Zero` to be more concise (and memory friendly)?
```scala
val initialGraceTime = {
  val now = System.currentTimeMillis()
  if(now >= deadline) Duration.Zero
  else (deadline - now).millis
}
```

e-orz: negative duration is handled ok for the Duration class and also the scheduler. I checked it before.
And for the very rare case, should we really be explicit?

hochgi: I'm not sure what'll happen here:
https://github.com/thomsonreuters/CM-Well/blob/master/server/cmwell-util/src/main/scala/cmwell/util/concurrent/SimpleScheduler.scala#L72
.toMillis returns 0? negative value?
timer handles both cases?
I prefer to know for sure and depend on implementation.
also, though not the case, explicitness may lead to less overhead in different APIs.
see: https://github.com/thomsonreuters/CM-Well/blob/master/server/cmwell-util/src/main/scala/cmwell/util/concurrent/package.scala#L214
for example

GuyGuterman:
Validated that iterator scheduler was working as expected. Tested by create iterators and used tokens for next chank thanks to the cmw-iterate tool. In the total was `~50K`. Then to analyze, memory has created a dump for all memory `(total 8.9GB)`. To analyzing this dump was used "memory analyzer" and found that system working as expected with iterators and each request worked and not catch to much time or stuck the queue.
Thank you for all the help @e-orz !

Tested and Verified
Moving to Done

PR #586, opened 2018-03-20
fixing several ingest issues

hochgi:
duplicate feedbacks. 
same inferred prefix to multiple namespaces in a single doc will fail now. and also I updated some dependencies. 

Some tests fail though... So still a WIP


server/cmwell-ws/app/ld/util/LDFormatParser.scala
```diff
@@ -862,42 +867,44 @@ object LDFormatParser extends LazyLogging {
...
-        //      TODO: replace current code with commented out, once all the old data is updated.
-        //      val t = CMWellRDFHelper
-        //        .urlToHash(url)
-        //        .map(_ -> Exists)
-        //        .getOrElse(CMWellRDFHelper.nsUrlToHash(url))
-        //      t -> url
-        cmwellRDFHelper.nsUrlToHash(url,timeContext) -> url
-      }.toSeq
+
+      it.foldLeft(Map.empty[String,(String,PrefixState)]) {
```
e-orz: use Map builder instead.

PR #599, opened 2018-03-26
Handle trailers from Chunked HttpResponse in Retry.scala and trigger …

nicklester:
…retry to _sp if X-CM-WELL-SG-RS header is present


server/cmwell-data-tools/src/main/scala/cmwell/tools/data/utils/akka/Retry.scala
```diff
@@ -186,18 +188,28 @@ object Retry extends DataToolsLogging with DataToolsConfig{
...
+                case (accumulatedEntity, chunk) => (accumulatedEntity._1 ++ chunk.data, None)
               }
             }
+            case entity:HttpEntity => {
+              entity.withoutSizeLimit.dataBytes.runFold(blank)(_ ++ _).map (_ -> None)
+            }
+          }).map {
+            case (entityBytes, trailerHeaders) =>
+
+              val combinedHeaders = headers ++ trailerHeaders.getOrElse(Seq.empty)
```
e-orz: maybe you should try this (more scala way):
`trailerHeaders.fold(headers)(headers ++ _)`
this way you won't create another object in case there are no footers.

server/cmwell-data-tools/src/main/scala/cmwell/tools/data/utils/akka/Retry.scala
```diff
@@ -186,18 +188,28 @@ object Retry extends DataToolsLogging with DataToolsConfig{
...
+
+              responseValidator(entityBytes, combinedHeaders) match {
```
e-orz: response validator hides the true reason why it failed.
maybe the function should get a response and return Try with Success(response) or Failure(failure reason)

It doesn't matter much how it will be implemented just I want the logs to be clear with the exact failure reason (unless it's written some other place)

server/cmwell-data-tools/src/main/scala/cmwell/tools/data/utils/akka/Retry.scala
nicklester: Yeah! Much nicer! Done

e-orz: If you want it more cryptic (and more scala style) you can do:
`trailerHeaders.fold(headers)(_.++)`

server/cmwell-data-tools/src/main/scala/cmwell/tools/data/utils/akka/Retry.scala
nicklester: What about function returns Try[Boolean], or maybe Either[Boolean, Seq[String]] (sequence of String being failure reasons...?

e-orz: Gilad likes Either, I less.
A Seq of failures is a good idea but I want somehow to expose it to the "outer" world (e.g. logs and such). Usually you have the ability to propagate an exception.
If you find a way to expose all the error it's better.
Try[Boolean] is the simplest but I prefer Try[Unit]. You won't use the value anyway.

server/cmwell-data-tools/src/main/scala/cmwell/tools/data/sparql/SparqlProcessor.scala
```diff
@@ -173,8 +174,23 @@ class SparqlProcessor[T](baseUrl: String,
...
-      !(body containsSlice "Could not process request")
+    def validateResponse(body: ByteString, headers: Seq[HttpHeader]) : Try[Unit] = {
+      def validateBody(body: ByteString) = {
+        if (!(body containsSlice "Could not process request")) None
+        else Some("Response body was not valid")
+      }
+
+      def validateHeaders(headers: Seq[HttpHeader])  = {
+        if(!(headers.contains("X-CM-WELL-SG-RS"))) None
+        else Some("Error reported in trailer headers")
```
e-orz: please expose the actual header (with the failed error numbers). Talked to @bryaakov and he said it will help analyzing errors.

nicklester: Agree. Implemented

PR #611, opened 2018-04-02
Fixed licenses in examples, added new python consumer example

TRnonodename:


TRnonodename:
Wonderful idea @hochgi Kicking myself that I didn't think of that.. changing now.

Issue #610, opened 2018-04-02
STP - Retry mechanism on failed materilizer should not be endless


yanpom:
Once the materialization of infoton fail it has a retry after 30 seconds – this retry seems to be endless and there is no limit of retries (at least that the info I’ve got from Eli/Nick)
Materialization should end after X number of retries and write the infoton to red.log (as part of 409) 
What happens if during the retries, manager is restarted? The concern is that data tools will lose the failed items and we’ll have no notification what had actually failed.

nicklester:
Have implemented configuration in reference.conf.

cmwell.sparql.http-retry-count

If this is not supplied then the _sp API will be called (as it worked before) repeatedly until success. If config is supplied one can specify an int value for the maximum number of requests to make. Should all requests fail then log will be written to bad data log. Writing to bad data log is existing functionality for the data tools app, but has been introduced as part of the PR #613. 

Is bad data log sufficient, or also write to red log as requested?

yanpom:
Not fixed

yanpom:
Verified and Closed

PR #614, opened 2018-04-05
#610 Retry mechanism on failed materilizer should not be endless

nicklester:


nicklester:
Yeah - this was done because I needed the prior change to test. Have separated this out now and forced a push... Hope that is ok.....

PR #622, opened 2018-04-11
Updated akka-http version

hochgi:
It contains a possible fix to the problem we saw with the previous release.
https://github.com/akka/akka-http/pull/1947
This should be validated by checking STP code against the new version.

@nicklester please test that STP runs correctly with the new version


e-orz:
I think it's not related. but checking with the legacy connection pool is something we want to do.

e-orz:
Just a thought: data-tools will always use theirs config with fallback to the normal akka config.
We talked about it before...

YoniMataraso:
We will handle this as part of our upgrade process, no reason to have it open for so long.

PR #624, opened 2018-04-11
Limit retries if errors in trailers

nicklester:


server/cmwell-data-tools/src/main/scala/cmwell/tools/data/utils/akka/Retry.scala
```diff
@@ -182,16 +182,12 @@ object Retry extends DataToolsLogging with DataToolsConfig {
...
-          logger.warn(
-            s"$labelValue received $s but response body is not valid, will retry again in $delay host=${getHostnameValue(h)} data=${stringifyData(data)}"
-          )
-          val future = after(delay, system.scheduler)(Future.successful(data))
-          Some(immutable.Seq(future -> state))
-
         case State(data, _, Some(HttpResponse(s, h, e, _)), count) =>
+
+          if (s.isSuccess)
+            logger.warn(s"$labelValue received $s but response body is not valid, will retry again in " +
```
e-orz: This log print will be even if there will be no retries and it'll confuse the one who will look at the logs.
cc @yanpom

PR #627, opened 2018-04-11
Utf8 charset

YoniMataraso:
Adding charset=utf-8 to request parameter


FlipGFlop:
Thanks for fixing this so quickly.

PR #640, opened 2018-04-17
STP performance changes..

nicklester:


nicklester:
Agree. Changes committed and pushed

server/cmwell-data-tools/src/main/scala/cmwell/tools/data/downloader/consumer/BufferFillerActor.scala
```diff
@@ -230,7 +235,8 @@ class BufferFillerActor(threshold: Int,
       val to = toHint.map("&to-hint=" + _).getOrElse("")
 
       val uri =
-        s"${formatHost(baseUrl)}/$consumeHandler?position=$token&format=tsv$paramsValue$slowBulk$to"
+        s"${formatHost(baseUrl)}/$consumeHandler?position=$token&format=tsv$paramsValue$slowBulk$to" +
+          consumeLengthHint.fold(""){"&length-hint=" + _ }
```
e-orz: by using length-hint along with bulk-consume you are actually ruining the optimizations done by the server to save ES searches.
e.g. The amount of searches will be DOUBLED!

e-orz: The BufferFillerActor is not only used for STP also for all the data tools. Even if STP will be configured to use only consume the data tools still can use bulk consume.

server/cmwell-data-tools/src/main/resources/reference.conf
```diff
@@ -20,6 +20,7 @@ cmwell {
   downloader {
     consumer.http-retry-timeout = "10.seconds"
     consumer.data-fetch-retries = 5
+    consumer.fetch-size = 100
```
e-orz: I think you meant to change the STP length-hint. Changing the config here will change the fetch size for all the data-tools. e.g. bulk-consume will stop to work completely (the server guards again too small chunk, or length-hint is not allowed for bulk-consume - don't remember)

Also the normal consumer should benefit from the 3000 chunk size of consume. Only the STP should be changed. Or am I missing something and everything got broke?

server/cmwell-data-tools/src/main/scala/cmwell/tools/data/downloader/consumer/BufferFillerActor.scala
nicklester: Ok - can get length-hint to run only for normal consume, no problem.
@YoniMataraso are you happy the length hint runs for all data tools, or do we need to work to limit it to STP only?

server/cmwell-data-tools/src/main/resources/reference.conf
nicklester: Correct @e-orz - my mistake. Not sure I understand what you mean about 'normal consumer'..

server/cmwell-sparql-agent/src/main/resources/reference.conf
```diff
@@ -10,5 +10,8 @@ cmwell.agents {
     init-delay = 10 seconds
     max-delay = 5 minutes
     config-polling-interval = 30 seconds
+    consumer {
```
e-orz: By manually adding this config to the STP agent, the STP  stand alone won't have the ability to change the length-hint.

You should:
Either
* add this feature the the stand alone STP 
* or revert back to the old commit and get the config from the application.conf of STP instead of the generic reference.con. This way you won't need to pass the parameter all along the way, just change the application.conf of the stand alone and the STP agent. It's your call but I prefer it this way - much cleaner and we will have this ability also for the consumer.

server/cmwell-sparql-agent/src/main/scala/cmwell/tools/data/sparql/SparqlProcessorManager.scala
```diff
@@ -422,7 +422,7 @@ class SparqlProcessorManager(settings: SparqlProcessorManagerSettings) extends A
     val hostUpdatesSource = job.config.hostUpdatesSource.getOrElse(settings.hostUpdatesSource)
 
     val agent = SparqlTriggeredProcessor
-      .listen(job.config, hostUpdatesSource, false, Some(tokenReporter), Some(job.name))
+      .listen(job.config, hostUpdatesSource, false, Some(tokenReporter), Some(job.name), consumeLengthHint=settings.consumeLengthHint)
```
e-orz: same argument as above - should we want this config to be STP **agent** only?

server/cmwell-sparql-agent/src/main/scala/cmwell/tools/data/sparql/SparqlProcessorManagerSettings.scala
```diff
@@ -40,6 +40,12 @@ class SparqlProcessorManagerSettings {
     stpSettings.getDuration("cmwell.agents.sparql-triggered-processor.max-delay").toMillis.millis
   val interval: FiniteDuration =
     stpSettings.getDuration("cmwell.agents.sparql-triggered-processor.config-polling-interval").toMillis.millis
+
+  val consumeLengthHint = stpSettings.hasPath("cmwell.agents.sparql-triggered-processor.consumer.fetch-size") match {
```
e-orz: same argument as above - should we want this config to be STP **agent** only?

server/cmwell-sparql-agent/src/main/resources/reference.conf
nicklester: @YoniMataraso - do we also want the stand alone to change length hint. I didn't think it was the case. If we decide I will have to change a few things in the PR

server/cmwell-data-tools/src/main/resources/reference.conf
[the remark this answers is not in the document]
nicklester: Agree done in commit #2b277a7

server/cmwell-data-tools/src/main/scala/cmwell/tools/data/downloader/consumer/BufferFillerActor.scala
```diff
@@ -230,7 +231,14 @@ class BufferFillerActor(threshold: Int,
       val to = toHint.map("&to-hint=" + _).getOrElse("")
 
       val uri =
-        s"${formatHost(baseUrl)}/$consumeHandler?position=$token&format=tsv$paramsValue$slowBulk$to"
+        s"${formatHost(baseUrl)}/$consumeHandler?position=$token&format=tsv$paramsValue$slowBulk$to" +
+         consumeLengthHint.fold(""){ hint =>
+            isBulk match {
```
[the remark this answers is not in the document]
nicklester: I don't know if the requirement is that even if using bulk and the system has degraded to consume we still want to respect the config... @YoniMataraso  - thoughts?

server/cmwell-sparql-agent/src/main/scala/cmwell/tools/data/sparql/SparqlProcessorManagerSettings.scala
YoniMataraso: Let's have a dedicated configuration for STP as @e-orz had suggested.

server/cmwell-data-tools/src/main/scala/cmwell/tools/data/downloader/consumer/BufferFillerActor.scala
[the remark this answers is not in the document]
YoniMataraso: Let's only use the length-hint in cases the consumer is non bulk one. If it is not specified it will use the default.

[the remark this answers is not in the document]
nicklester: Agreed. Won't change this then

server/cmwell-sparql-agent/src/main/resources/reference.conf
nicklester: Will create stand alone data tools changes in a future PR after discussion with @YoniMataraso

server/cmwell-sparql-agent/src/main/scala/cmwell/tools/data/sparql/SparqlProcessorManagerSettings.scala
nicklester: Will create stand alone data tools changes in a future PR after discussion with @YoniMataraso

PR #651, opened 2018-04-24
Not using Preferred BgActor but 3 possible (adjacent) hosts

bryaakov:
will solve #654

### DIFF
--- a/server/cmwell-bg/src/main/scala/cmwell/bg/CMWellBGActor.scala
+++ b/server/cmwell-bg/src/main/scala/cmwell/bg/CMWellBGActor.scala
@@ -19,6 +19,8 @@ import java.util.concurrent.TimeUnit
 import akka.actor.{Actor, ActorRef, OneForOneStrategy, PoisonPill, Props}
 import akka.kafka.{ConsumerSettings, KafkaConsumerActor}
 import akka.stream.{ActorMaterializer, Supervision}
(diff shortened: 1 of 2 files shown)


server/cmwell-bg/src/main/scala/cmwell/bg/CMWellBGActor.scala
```diff
@@ -84,6 +89,20 @@ class CMWellBGActor(partition: Int,
     logger.info(s"CMwellBGActor-$partition starting")
     super.preStart()
     self ! Start
+
+    import akka.pattern.ask
+    sys.addShutdownHook {
```
e-orz: consider this scenario:
on a particular JVM:
1. BGActor5 is run on it.
2. sys.addShutdownHook {BActor5 ! shutDown}
3. BGActor5 crashes - no one runs the shutdown???
4. BGActor5 starts on another JVM
5. the original JVM crashes - BGActor5 is shutdown!!!

e-orz: Do you see any reason why no to add the shutdown logic to the postStop of the BGActor?

bryaakov: `postStop` sounds like a better plan. Will that answer the scenario you described?

PR #659, opened 2018-04-30
Matching font size in searchbox to search-where dropdown

bryaakov:


AnnaGrebnev:
@yanpom  please check new UI issue

bryaakov:
@yanpom this was a change request from UX team, to match the Search Box font size with the "search where" dropdown font size. It used to be 20px for the search and 15px for the dropdown, now both are 15.

PR #672, opened 2018-05-07
Storing Null Updates in zStore

bryaakov:


server/cmwell-bg/src/main/scala/cmwell/bg/ImpStream.scala
```diff
@@ -687,11 +678,19 @@ class ImpStream(partition: Int,
                   case (reason, None) => Future(logger .info (s"evicted command in path:$path because of: $reason"))
                 }
               }
+              val zStored = {
+                val ttlSeconds = 7.days.toSeconds.toInt //TODO propagate offsets.retention.minutes's value to here
+                travector(offsets.map(_.offset)) { o =>
+                  val (key, payload) = s"bg.offset_$o" -> "nu".getBytes(StandardCharsets.UTF_8)
```
e-orz: offset is a per-partition info and not general.

PR #680, opened 2018-05-12
Bumping Scala to 2.12.6

schrepfler:
Updating as 2.12.5 had issues with macros when compiling on Java 9 or 10.

### DIFF
--- a/server/build.sbt
+++ b/server/build.sbt
@@ -20,7 +20,7 @@ sys.env.get("BUILD_NUMBER") match {
   )
 }
 
-scalaVersion in Global := "2.12.5"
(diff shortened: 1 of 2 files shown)


schrepfler:
I think I've just ran it, haven't done any particular tests.

YoniMataraso:
Hi @schrepfler - upgrading a Scala version - even if it's minor requires us to verify that CM-Well can be compile/build and deployed on a cluster before we can make this change.

We are planing to upgrade our libraries including Scala in the future. If you want to help by doing some of the tests you can start by running sbt ccft and continue from there.

For now I'm closing this PR.

PR #692, opened 2018-05-21
STP token and statistics now saved to zStore

nicklester:


server/cmwell-sparql-agent/src/main/scala/cmwell/tools/data/sparql/InfotonReporter.scala
```diff
@@ -131,38 +128,37 @@ class InfotonReporter private (baseUrl: String, path: String)(implicit mat: Mate
...
-          logger.error(s"problem writing tokens infoton to $path")
-          e.discardBytes()
+      .map{createTokenPayload}
+      .map{
+        zStore.putString(s"stp-agent-${extractLastPart(path)}", _)
       }
+      .map(zStoreFuture => {
+        zStoreFuture.onComplete {
+          case Success(_) => logger.debug(s"successfully written tokens to zStore, key=$path")
+          case Failure(ex) => logger.error(s"problem writing tokens to zStore: ${ex.getMessage}")
```
bryaakov: it is better to `logger.error("explanation...", ex)` - in that case you have all StackTrace.

server/cmwell-sparql-agent/src/main/scala/cmwell/tools/data/sparql/StpUtil.scala
```diff
@@ -31,42 +40,69 @@ object StpUtil {
...
+    def shouldRetry(action: String): (Try[TokenAndStatisticsMap], RetryParams) => ShouldRetry[RetryParams] = {
+      import scala.language.implicitConversions
+      implicit def asFiniteDuration(d: Duration) = scala.concurrent.duration.Duration.fromNanos(d.toNanos);
+      {
+        case (Success(_),_) =>
+          logger.debug(s"Successfully read token and statistics state from zStore for agent ${extractLastPart(path)}")
+          DoNotRetry
+        case (Failure(ex), state) if state.retriesLeft == 0 => {
+          logger.error(ex.getMessage)
+          logger.error(s"Failed to read token and statistics state from zStore. 0 retries left. Will not re-attempt for agent ${extractLastPart(path)}")
```
bryaakov: I'd combine these two logger.error lines into one, first the s"Faild ..." string and then `, ex)` (without .getMessage)

Issue #655, opened 2018-04-29
httpParallelism change in PR #640 only changes consuming the dataBytes and not the total requests


e-orz:
look at:
https://github.com/nicklester/CM-Well/blob/577a6bd2af539c2b6ab62ed39a6bf693ada75542/server/cmwell-data-tools/src/main/scala/cmwell/tools/data/utils/akka/Retry.scala#L278
which is still same as before.
Also note
https://github.com/nicklester/CM-Well/blob/577a6bd2af539c2b6ab62ed39a6bf693ada75542/server/cmwell-data-tools/src/main/scala/cmwell/tools/data/utils/akka/Retry.scala#L277
The connection pool acts like a buffer and has internal parallelism in it.
How many connection? I think that it is according to the config.

e-orz:
It can be fixed by giving the data-tools special akka config section (very easy to do).
It can help also with PR #622

nicklester:
As discussed wont modify:
.mapAsyncUnordered(httpParallelism) { 

due to the issue:

- Retry.scala - the retryHttp method takes a parallelism parameter, however it is not used - in favour of a function of two akka params:  val maxConnections =
     config.getInt("akka.http.host-connection-pool.max-connections")
   val httpPipelineSize =
     config.getInt("akka.http.host-connection-pool.pipelining-limit")
   val httpParallelism = maxConnections * httpPipelineSize

However I have introduced the akka config.

nicklester:
After discussion with @e-orz we have concluded that we need to extract all data tools config into a data tools specific typesafe config, so that we don't affect general dc config etc when tuning STP settings. The change is fairly straightforward, but may require significant testing

nicklester:
@e-orz Does PR #747 cover this? Retry.scala is now using max connections and pipeline limit from the new data-tools typesafe config

e-orz:
@nicklester It covers it but only partially. Now the connection pool has its own config taken from the data-tools section and this config sets the connection pool internal buffer/parallelism.
The httpParallelism is used only for the processing after the responses got from the pool.

Bottom line: I think this issue can be closed be be careful when configuring STP and set all the confgs with proper values.

PR #702, opened 2018-05-25
Updated Akka Http parrallelisation config

nicklester:
Referencing issue #655


server/cmwell-data-tools/src/main/resources/reference.conf
```diff
@@ -10,6 +10,7 @@ akka {
   logging-filter = "akka.event.slf4j.Slf4jLoggingFilter"
 
   http.host-connection-pool {
+    max-connections = 10
```
e-orz: in the dc process it's overridden in application.conf and this config will be ignored for the STP agent mode.
The correct way to solve it is to have a dedicated config section for the data-tools (as I already wrote in the past).

YoniMataraso:
It will not work as expected with the current config.

PR #706, opened 2018-05-28
Decrease tests running time | AuthTests

MoriaYechiel:


server/cmwell-it/src/it/scala/cmwell/it/AuthTests.scala
```diff
@@ -142,16 +143,18 @@ class AuthTests extends FunSpec with Matchers with Helpers with LazyLogging {
...
-          val res = waitAndExtractBody(Http.post(path, exampleObj, Some("application/json"), headers = ("X-CM-Well-Type" -> "File") :: tokenHeader))
-          Json.parse(res) should be(jsonSuccess)
+        ) map { path =>
+          waitAndExtractBody(Http.post(path, exampleObj, Some("application/json"), headers = ("X-CM-Well-Type" -> "File") :: tokenHeader))
         }
-        indexingDuration.fromNow.block
+      }
+
+      it("should add some example data") {
+        exampleData.map(Json.parse(_) should be(jsonSuccess))
```
e-orz: very nice separation of the tests!
1. As a convention if the result is not needed you should use `foreach`
2. It's better to use the scalatest way. e.g. `all (exampleData.map(Json.parse)) should be(jsonSuccess)`

server/cmwell-it/src/it/scala/cmwell/it/AuthTests.scala
```diff
@@ -142,16 +143,18 @@ class AuthTests extends FunSpec with Matchers with Helpers with LazyLogging {
...
       }
 
       it("should reset AuthCache once all user/role data was ingested") {
-        val res = waitAndExtractBody(Http.get(cmw / "_auth", List("op" -> "invalidate-cache"), tokenHeader))
-        Json.parse(res) should be(jsonSuccess)
+        spinCheck(100.millis, true)(Http.get(cmw / "_auth", List("op" -> "invalidate-cache"), tokenHeader))( res => Json.parse(res.payload).equals(jsonSuccess))
```
e-orz: in scala you should use `==` and not `equals`

PR #713, opened 2018-06-03
updating some dependencies

hochgi:


YoniMataraso:
Thank you @hochgi for this Pull Request,
It contains too many library updates at the same time. We will soon share a plan of planned library upgrades.

If you don't mind we are going to take your SBT changes and issue a new PR just with them.
SBT will be upgraded in our next drop/version.

PR #714, opened 2018-06-03
Adding Get API to FTS

bryaakov:


bryaakov:
Oops! Will expose that as well. Stand by for one more commit...

PR #728, opened 2018-06-10
Fix java_home to be calculated on remote side

MoriaYechiel:


server/cmwell-ws/app/controllers/Health.scala
```diff
@@ -119,9 +116,10 @@ class Health @Inject()(crudServiceFS: CRUDServiceFS, ws: WSClient) extends Injec
 
   def getKafkaStatus = Action.async {implicit req =>
     val javaHomeLocation = s"$path/../java"
-    val javaHomeAddition = if (Files.exists(Paths.get(javaHomeLocation)))
-      s"JAVA_HOME=$path/../java"
-    else ""
+
+    val javaHomeAddition = s"""if [ -d $javaHomeLocation ]
+        then JAVA_HOME=$javaHomeLocation ;
```
e-orz: this won't work and you will need export (even though that without the "if" it will work)
(at least without the bash -c it doesn't work)

server/cmwell-cons/src/main/scala/ctl.scala
```diff
@@ -1615,9 +1615,10 @@ abstract class Host(user: String,
     val replicationFactor = math.min(hosts.size, 3)
 
     val javaHomeLocation = s"${instDirs.globalLocation}/cm-well/app/java"
-    val exportCommand = if (Files.exists(Paths.get(javaHomeLocation)))
-      s"export PATH=${instDirs.globalLocation}/cm-well/app/java/bin:$$PATH ; export JAVA_HOME=${javaHomeLocation} ;"
-      else ""
+    val exportCommand = s"""if [ -d $javaHomeLocation ]
+        then export PATH=$javaHomeLocation/cm-well/app/java/bin:$$PATH ;
```
e-orz: this is a type - you need to add only `bin` and not the whole path

server/cmwell-cons/src/main/scala/ctl.scala
```diff
@@ -1874,9 +1875,12 @@ abstract class Host(user: String,
     val replicationFactor = math.min(ips.size, 3)
 
     val javaHomeLocation = s"${absPath(instDirs.globalLocation)}/cm-well/app/java"
-    val exportCommand = if (Files.exists(Paths.get(javaHomeLocation)))
-      s"export PATH=${absPath(instDirs.globalLocation)}/cm-well/app/java/bin:$$PATH ; export JAVA_HOME=${javaHomeLocation} ;"
-    else ""
+
+    val exportCommand = s"""if [ -d $javaHomeLocation ]
+        then export PATH=$javaHomeLocation/cm-well/app/java/bin:$$PATH ;
```
e-orz: same as above

PR #743, opened 2018-06-14
Expose Kafka topic in ws

### LINKED ISSUE #744: Add Kafka Topic Consuming API in WS

bryaakov:
fix #744

### DIFF
--- a/server/cmwell-ws/app/controllers/Application.scala
+++ b/server/cmwell-ws/app/controllers/Application.scala
@@ -17,25 +17,19 @@ package controllers
 import java.net.URLEncoder
 import java.nio.charset.StandardCharsets
 import java.util.Locale
-import javax.inject._
 
+import javax.inject._
 import actions._
 import akka.NotUsed
 import akka.actor.ActorRef
+import akka.kafka.{ConsumerSettings, Subscriptions}
+import akka.kafka.scaladsl.Consumer
 import akka.pattern.AskTimeoutException
 import akka.stream.scaladsl.Source
 import akka.util.{ByteString, Timeout}
+import cmwell.common.CommandSerializer
 import cmwell.common.file.MimeTypeIdentifier
-import cmwell.domain.{
-  BagOfInfotons,
-  CompoundInfoton,
-  DeletedInfoton,
-  FString,
-  PaginationInfo,
-  SearchResponse,
-  SearchResults,
-  _
-}
+import cmwell.domain.{BagOfInfotons, CompoundInfoton, DeletedInfoton, FString, PaginationInfo, SearchResponse, SearchResults, _}
 import cmwell.formats._
 import cmwell.fts._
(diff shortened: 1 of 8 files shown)


server/cmwell-ws/app/logic/CRUDServiceFS.scala
```diff
@@ -462,6 +469,19 @@ class CRUDServiceFS @Inject()(implicit ec: ExecutionContext, sys: ActorSystem) e
     }
   }
 
+  def consumeKafka(topic: String, partition: Int, offset: Long, maxLengthOpt: Option[Long]): Source[Array[Byte], Consumer.Control] = {
+    val byteArrayDeserializer = new ByteArrayDeserializer()
+    val subscription = Subscriptions.assignmentWithOffset(new TopicPartition(topic, partition) -> offset)
+    val consumerSettings = ConsumerSettings(Grid.system, byteArrayDeserializer, byteArrayDeserializer)
+      .withBootstrapServers("localhost:9092")
```
e-orz: You shouldn't hard-code the bootstrap servers (and in a clustered environment it should have 3 and not just one)

bryaakov: So if this will happen to reach a WS on one of non “seed nodes”, will it fail? I’m going to have to wire that in cons... :/

server/cmwell-ws/app/controllers/Application.scala
```diff
@@ -3004,6 +3001,27 @@ callback=< [URL] >
...
+    if (sysTopics(topic) && !authUtils.isOperationAllowedForUser(Admin, authUtils.extractTokenFrom(req), evenForNonProdEnv = true))
+      Forbidden("Consuming this topic requires an Admin token.")
+    else {
+      val offset = req.getQueryString("offset").fold(0L)(_.toLong)
+      val maxLengthOpt = req.getQueryString("max-length").map(_.toLong)
+      val isText = req.getQueryString("format").fold(false)(_ == "text")
+
+      val source = crudServiceFS.consumeKafka(topic, partition, offset, maxLengthOpt)
+      Ok.chunked(source.map { bytes =>
+        (Try(CommandSerializer.decode(bytes)) match {
```
e-orz: For the Crawler topic it will fail (throw an exception???) for each message and only then will return the actual message.

Also for the crawler topic it will get into:
https://github.com/thomsonreuters/CM-Well/blob/04039b236a43dc2e86990a80d7cf97e0946b545d/server/cmwell-common/src/main/scala/cmwell/common/formats/JsonSerializer.scala#L133
And will throw an exception in the server.

bryaakov: won't it get to that line: `case Failure(_) if isText => new String(bytes, StandardCharsets.UTF_8)` ?

e-orz: might be but it will throw an exception for **each** element!

PR #783, opened 2018-06-25
Crawler persists into red_queue

bryaakov:


bryaakov:
Closing, and creating a new one with same content based on updated master.

PR #792, opened 2018-07-01
Hide asterisk in Rsync command

MoriaYechiel:


server/cmwell-cons/src/main/scala/ctl.scala
```diff
@@ -789,7 +789,7 @@ abstract class Host(user: String,
   }
 
   def _rsync(from: String, to: String, host: String, tries: Int = 10, sudo: Boolean): Try[String] = {
-    val seq = Seq("rsync", "-Paz", "--delete", from, host + ":" + to)
+    val seq = Seq("rsync", "-e", "'ssh -o LogLevel=ERROR'", "-Paz", "--delete", from, host + ":" + to)
```
e-orz: remove the \`s

PR #817, opened 2018-07-11
Add tools for data consistency analysis

Andy-42:
These are the data consistency analysis tools that were previously in my personal repository https://github.com/Andy-42/CM-Well. I've created a new top-level folder **analysis** to contain these two projects since this doesn't belong in the server project.


Andy-42:
OK. I've moved the tools.

PR #831, opened 2018-07-19
Add logback-test to ws

MoriaYechiel:


server/cmwell-ws/src/test/logback-test.xml
```diff
@@ -0,0 +1,134 @@
+<configuration debug="false">
+
```
e-orz: add also the console appender, as you did with all the other logback-test files.

PR #864, opened 2018-08-02
851 : add upper limit to length-hint consume API.

YoniMataraso:


server/cmwell-ws/app/controllers/Application.scala
```diff
@@ -1470,7 +1470,7 @@ callback=< [URL] >
                 if (isSimpleConsume) Settings.consumeSimpleChunkSize
                 else Settings.consumeExpandableChunkSize
               }
-          )
+          ).min(Settings.consumeSimpleChunkSize)  //Making sure length hint is not bigger than max
```
bryaakov: How do you know if this is the relevant value? Perhaps it is `Settings.consumeExpandableChunkSize`?
May I suggest to factor out the `if ... else` from lines 1470, 1471 to a val before line 1459, e.g. `val settingsChunkSize = if(isSimpleConsume) etc.` - then, same one can be used both in `.getOrElse` and in `min`.

YoniMataraso: @bryaakov I'm looking for the min(your value, Settings.consumeExpandableChunkSize->3000)

as I didn't want to go and actually hardcode the number 3K.
Let me know if you see any issue with it.

bryaakov:
See PR #1008

PR #885, opened 2018-08-17
Fix consistency threshold

Andy-42:
Correct implementation of consistency threshold. The previous implementation didn't make sense, and could return false positives when run on a system actively ingesting data.

Add logging to show the consistency threshold used for analysis.

Rename --current-threshold parameter to --consistency-threshold since the parameter's format is changing, and the original name was poorly chosen.


bryaakov:
There were conflicts, I'm closing this one and creating a new one with the conflicts resolved.

PR #908, opened 2018-09-02
Persist plusOne in zStore

bryaakov:


server/cmwell-bg/src/main/scala/cmwell/bg/ImpStream.scala
```diff
@@ -300,6 +300,19 @@ class ImpStream(partition: Int,
         bgMessage.copy(message = (baseInfoton -> mergedInfoton))
       }
   }
+
+  val persistNewLastModified = Flow[BGMessage[(Option[Infoton],MergeResponse)]].mapAsync(irwWriteConcurrency) {
+    case bm@BGMessage(offsets, (_, mr)) => mr.extra.fold(Future.successful(bm)) { extra =>
+      val ttlSeconds = 7.days.toSeconds.toInt //TODO propagate offsets.retention.minutes's value to here
+      travector(offsets) { o =>
```
e-orz: All the "inner" offsets are "grp" only the one that is not considered grouped should be added to zStroe.

PR #909, opened 2018-09-02
Basic flow of upgrade

MoriaYechiel:


server/cmwell-cons/src/main/scala/ctl.scala
```diff
@@ -2005,6 +2009,8 @@ abstract class Host(user: String,
               withUpdateSchemas: Boolean = false,
               hosts: GenSeq[String] = ips) {
 
+    val currentVersion = extractVersionFromProcNode(ips(0))
+    currentVersion.map(ver => info(s"Current version is $ver"))
```
bryaakov: What if we failed to fetch /proc/node or the JSON parsing failed for some reason?
We will just move on with upgrade. I recommend in this case to use `Await.result` even it's not considered as a best practice, we don't want to proceed with upgrade if we don't know the current version. 
(In future we will refactor the whole thing to be less imperative.)

server/cmwell-cons/src/main/scala/ctl.scala
```diff
@@ -2174,6 +2180,25 @@ abstract class Host(user: String,
...
+    Upgrade.runPostUpgradeActions(currentVersion, upgradedVersion)
+    info(s"Upgrade completed!")
+  }
+
+  def extractVersionFromCmwellProperties : String = {
+    val cmwellProp = Source.fromURL(this.getClass.getResource(cmwellPropertiesFile)).mkString
+    (Json.parse(cmwellProp) \ "cm-well_version").as[String].replace("x-SNAPSHOT", "0")
+  }
+
+  def extractVersionFromProcNode(host : String) : Future[String] = {
```
bryaakov: You can wrap the entire function body with `cmwell.util.concurrent.retry(...) { ... }`

PR #914, opened 2018-09-02
Upgrade logic

MoriaYechiel:


server/cmwell-cons/src/main/scala/Upgrade.scala
```diff
@@ -0,0 +1,56 @@
...
+
+import semverfi.{SemVersion, Version}
+
+import scala.collection.GenSeq
+import scala.concurrent.Future
+import scala.concurrent.ExecutionContext.Implicits.global
+
+object Upgrade {
+
+  class UpgradeFunc (desc: String, func: (GenSeq[String]) => Boolean){
```
bryaakov: use case class.

bryaakov: func should return `Future[Boolean]`, in line 25 you can wrap `func(hosts)` with an Await.result(..., 1.minute)

server/cmwell-cons/src/main/scala/Upgrade.scala
```diff
@@ -0,0 +1,56 @@
...
+    def executeUpgrade(hosts: GenSeq[String]) = if (func(hosts)) info(s"$desc succeeded") else error(s"$desc failed")
+  }
+
+  val postUpgradeMap : Map[String, UpgradeFunc] = {
+    //example
+    /*Map( "1.6.2" -> new UpgradeFunc("myUpgradeSomething",(hosts: GenSeq[String]) => {info(s"in func print: ${hosts(0)}"); true}),
```
bryaakov: Please include the version within the case class. e.g. `case class UpgradeFunc(version: SemVersion, desc: String, func: (GenSeq[String]) => Boolean)`
then, instead of `postUpgradeMap.foreach`, you can do `postUpgradeMap.sortBy(_.version).foreach`

server/cmwell-cons/src/main/scala/Upgrade.scala
```diff
@@ -0,0 +1,56 @@
...
+      "1.6.7" -> new UpgradeFunc("myUpgradeSomethingFailed",(hosts: GenSeq[String]) => {info(s"in func print failed: ${hosts(0)}"); false}))*/
+  }
+
+  def runPostUpgradeActions(currentVersionF : Future[String], upgraded : String, hosts: GenSeq[String])= {
+    def shouldRun(test : SemVersion, curr : SemVersion, upgrade : SemVersion) : Boolean = (test > curr) && (test <= upgrade)
+
+    currentVersionF.map(current => postUpgradeMap.foreach{f =>
```
bryaakov: instead of foreach and an if inside, you can
`val relevant = postUpgradeMap.filter(shouldRun)`
Then, please just before `relevant.foreach(_.func.execute)` logger.info/println s"Going to execute ${relevent.length} post-upgrade tasks"

PR #930, opened 2018-09-21
If server closes connection on _out , split into multiple requests

nicklester:


server/cmwell-data-tools/src/main/scala/cmwell/tools/data/utils/akka/Retry.scala
```diff
@@ -273,12 +273,36 @@ object Retry extends DataToolsLogging with DataToolsConfig {
         case State(data, _, None, count, _) =>
           count match {
             case Some(c) if c > 0 =>
-              logger.warn(
-                s"$labelValue error: could not send http request, counter=$c will retry again in $delay data=${stringifyData(data)}"
-              )
-              val future =
-                after(delay, system.scheduler)(Future.successful(data))
-              Some(immutable.Seq(future -> state.copy(count = Some(c - 1))))
+              if(state.data.size > 1) {
```
e-orz: Is None the only case that needs splitting? What about server error (I think client error is already handled)

AnnaGrebnev:
@yanpom Please check the issue

PR #961, opened 2018-10-28
Cluster Default RDF Protocol

bryaakov:


server/cmwell-common/src/main/resource/reference.conf
```diff
@@ -1,4 +1,8 @@
 metrics.reportMetricsJMX=true
 metrics.reportMetricsGraphite=false
 metrics.reportMetricsSlf4j=false
-metrics.graphite.host=vstat.clearforest.com
\ No newline at end of file
+metrics.graphite.host=vstat.clearforest.com
```
e-orz: remove the file from the PR

PR #969, opened 2018-10-30
Documentation updated for the new protocol system field, and handling…

nruppin:
… of the https value.


YoniMataraso:
@nruppin - can you please implements Yaakov's changes

PR #987, opened 2018-11-15
Pause, Restart and Reset Tokens buttons

nicklester:
root user access only...


server/cmwell-ws/conf/routes
```diff
@@ -61,6 +61,8 @@ GET         /ws/_out                controllers.OutputHandler.handleWebSocket(fo
 GET         /ii/*uuid               controllers.Application.handleUuidGET(uuid)
 GET         /zz                     controllers.Help.handleHelp(page = "zz")
 GET         /zz/*key                controllers.Application.handleZzGET(key)
+GET         /stp/*agent             controllers.Application.handleStpControl(agent)
```
bryaakov: please rename route to `/_stp/*agent`

server/cmwell-ws/app/controllers/Application.scala
```diff
@@ -3001,6 +3001,35 @@ callback=< [URL] >
     }
   }
 
+  def handleStpControl(agent: String) = Action.async(parse.raw) { implicit req =>
+
+    isAdminEvenNonProd(req) match {
+      case true => {
+
+        val headers = req.cookies.get("X-CM-WELL-TOKEN2") match {
```
bryaakov: Please use `AuthUtils.extractTokenFrom(req)` instead of getting the cookie explicitly.

server/cmwell-ws/app/controllers/Application.scala
```diff
@@ -3001,6 +3001,35 @@ callback=< [URL] >
...
+          case Some(cookie) => Seq(("X-CM-WELL-TOKEN", cookie.value))
+          case _ => Nil
+        }
+
+        val flag = req.getQueryString("enabled").flatMap(asBoolean).getOrElse(true)
```
bryaakov: I don't like the fact we are making a _in post from Application.scala, where we can simply invoke a `crudServiceFS` method... @nicklester, how complicated will it be to do that instead?

server/cmwell-ws/app/controllers/Application.scala
nicklester: Now redundant as not needed when using crudServiceFS

server/cmwell-ws/app/controllers/Application.scala
nicklester: Have done this - please can you take a look (FWIW - I completely agree 👍 )

nicklester: @bryaakov made another push..

server/cmwell-ws/conf/routes
nicklester: @bryaakov Done...

PR #993, opened 2018-11-21
729 - Option to use quads in _sp for STP

nicklester:


bryaakov:
A question about the architecture of this feature: Do we want that as a config setting per stp instance (which implies cluster wide setting), or perhaps a higher resolution of **per sensor** setting?

PR #997, opened 2018-11-26
Adding protocol system field Elastic mappings

bryaakov:


AnnaGrebnev:
@Ivan-Shestakov to add automation tests once it's done

PR #1004, opened 2018-11-27
Simple fix to display error on dashboard when sparql_materializer inf…

nicklester:
…oton is not found


nicklester:
@bryaakov - agree. Committed your suggestion. Good spot - thanks

Issue #1027, opened 2019-01-22
Protect CM-Well from CRLF Injection


bryaakov:
As was suggested by VeraCode, log lines should never contain user input, and it is recommended to sanitize all strings that are passed into any `logger.info` (or debug, trace, error, etc.) method.

For now there's a temp. solution in PR, but we need to have either

1. a wrapper for `LazyLogging`,
or,
2. some log4j Filter that does that for us

PR #1028, opened 2019-01-22
CRLF Injection Protection by Sanitization, as recommended by VeraCode

bryaakov:
This is a quick fix. We need to solve it more elegantly, as suggested in issue #1027.


YuriiYarosh:
I'll add some basic logback filter instead of `sanitizeLogLine` method, which should be much more convenient.

e-orz:
Make sure that your code is not just more convenient but also "passes" VeraCode restrictions. Without it, it won't worth much.

PR #1045, opened 2019-02-11
Doc updates mkdocs migration

nruppin:
Overhaul of all CM-Well Markdown documentation files and structure, for display with mkdocs.
Fixed all broken links.


e-orz:
Where is the mkdocs.yml file?

lielShraga:
@Ivan-Shestakov , we reviewed the code and saw that there is no usage for length in next-chunk api. basically, the length is determined in the create-iterator only.

PR #1083, opened 2019-03-24
Adding internal maven repo support

bryaakov:


e-orz:
It's fixed with PR #1087

PR #1087, opened 2019-03-31
Update coursier

e-orz:


server/cmwell-cons/build.sbt
```diff
@@ -32,21 +32,36 @@ getExternalComponents := {
   val logger = streams.value
   val dm = dependenciesManager.value
 
+/*
```
lielShraga: please remove comments

e-orz: done

PR #1090, opened 2019-04-01
delete ssh pass after install

lielShraga:


server/cmwell-cons/src/main/scala/ctl.scala
```diff
@@ -1127,7 +1127,9 @@ abstract class Host(user: String,
   }
 
   private def deleteSshpass(hosts: GenSeq[String], sudoer: Credentials): Unit = {
-    command("sudo rm sshpass", hosts, true, Some(sudoer))
+    info("delete ssh pass")
+    command("sudo rm ~/bin/sshpass", hosts, true, Some(sudoer))
```
e-orz: remove `sudo`

PR #1082, opened 2019-03-24
Install 1 cas and 4 es on the node ip

lielShraga:

PR #1096, opened 2019-04-03
expose use commit log as param

lielShraga:


bryaakov:
How does that differ from PR #1082 ?

lielShraga:
As discussed with Dudi we need to expose the casUseCommitLog param to non-oss conf file like other params that we expose.
But, we closed that I will give a default value in the code, since we want to keep the conf file small.(this is my last commit on this FB)

PR #1120, opened 2019-05-13
Atom release notes.

nruppin:


nruppin:
Dudi, I didn't forget about the 2 new config parameters. Moria says that none of the other config parameters are currently documented and we need to work on documenting the entire set of parameters together.

docs/ReleaseNotes/ReleaseNotes.Atom.May.2019.md
```diff
@@ -0,0 +1,32 @@
+# Atom Release (May 2019)
+
+!!! warning
+	- Due to significant data format changes, you must perform a full install of the Atom version. You cannot perform an upgrade process and retain your existing data. 
+	- To copy over your data to a CM-Well node with the new version, either re-ingest your data, or perform a DC-Sync process between the "old node" and the "new node".
```
bryaakov: Please replace "node" with "cluster". Moreover, we need to emphasize the need of a temp. cluster in the case of DC-Sync. It's good to state the the preferable option is re-ingest, but in cases when users cannot afford that, they need a temp. cluster like so: let's say A is the existing Zerba Cluster. and T is a temp. cluster with either Zebra or Atom but no data. Steps are: 1. DC-Sync A->T. 2. "wipe-and-install" Atom on A. 3. DC-Sync T->A.

docs/ReleaseNotes/ReleaseNotes.Atom.May.2019.md
```diff
@@ -0,0 +1,32 @@
+# Atom Release (May 2019)
+
+!!! warning
+	- Due to significant data format changes, you must perform a full install of the Atom version. You cannot perform an upgrade process and retain your existing data. 
+	- To copy over your data to a CM-Well node with the new version, either re-ingest your data, or perform a DC-Sync process between the "old node" and the "new node". 
+	- After installing Atom, you will be able to perform an upgrade process for subsequent releases.
+
+Title | Git Issue | Description 
+:------|:----------|:------------
+Major version upgrades to 3rd-party libraries | [1003](https://github.com/thomsonreuters/CM-Well/pull/1003), [1109](https://github.com/thomsonreuters/CM-Well/pull/1109) | Several 3rd-party libraries have been upgraded, including: Elasticsearch -> 7, Cassandra -> 3.11.4, Kafka -> 2.1.1. **NOTE that due to the significant data format changes caused by these upgrades, you must perform a full reinstall of CM-Well; you CANNOT perform an upgrade process that retains the data.**
```
bryaakov: do we need that note here, given the one above?

docs/ReleaseNotes/ReleaseNotes.Atom.May.2019.md
```diff
@@ -0,0 +1,32 @@
+# Atom Release (May 2019)
+
+!!! warning
```
MoriaYechiel: Looks like the display of the text is not working well here. If you meant for dots in the start of each line, it is not displayed. Also, what is '!!!' ? I think it will be better to highlight it with ** warning **

bryaakov: @MoriaYechiel that's fine; while the markdown preview in GitHub has some features, when we publish docs using mkDocs to github.io it turns out to look and feel just great.
_/* BTW: great minds think alike: I asked Eli if that `!!!` syntax is ok and he confirmed */_

MoriaYechiel: Cool! Didn't know

docs/ReleaseNotes/ReleaseNotes.Atom.May.2019.md
nruppin: Added.

docs/ReleaseNotes/ReleaseNotes.Atom.May.2019.md
nruppin: I shortened this note.

nruppin:
Please see updated version of release notes.

Issue #1121, opened 2019-05-13
WikiData Import: Illegal Subject Name


r-clancy:
I'm trying to import a number of triples from Wikidata and have come across the following error:

```
ryan@thinkpad ~/Downloads $ curl -X POST "http://localhost:9000/_in?format=ntriples" --data-binary @Q76.nt
{"success":false,"error":"illegal subject name: pt.wikinews.org/wiki/Categoria:Barack_Obama"}⏎       
                                                                                                                                                        
ryan@thinkpad ~/Downloads $ grep "pt.wikinews.org/wiki/Categoria:Barack_Obama" Q76.nt
<https://pt.wikinews.org/wiki/Categoria:Barack_Obama> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://schema.org/Article> .
<https://pt.wikinews.org/wiki/Categoria:Barack_Obama> <http://schema.org/about> <http://www.wikidata.org/entity/Q76> .
<https://pt.wikinews.org/wiki/Categoria:Barack_Obama> <http://schema.org/inLanguage> "pt" .
<https://pt.wikinews.org/wiki/Categoria:Barack_Obama> <http://schema.org/isPartOf> <https://pt.wikinews.org/> .
<https://pt.wikinews.org/wiki/Categoria:Barack_Obama> <http://schema.org/name> "Categoria:Barack Obama"@pt .
```

What about this subject name is invalid? Should the constraints on subject names be relaxed (see https://github.com/CM-Well/CM-Well/issues/156)? What kind of pre-processing do I need to do to the data?

bryaakov:
Looks like CM-Well doesn't like the colon character in Subject URIs.
- @r-clancy - for the time being, as a workaround, you can `sed 's/Categoria:/Categoria_/'` (if it won't hurt your use-case too much).
- @dudi3001 - I think we can allow that one, though I'd be careful not allow every character. What do you think?

r-clancy:
Thanks @bryaakov 

I'd vote for allowing the colon, the entire Wikidata `.nt` dump is about 1.3TB so it would be a lot of pre-processing.

bryaakov:
Have we completed testing this item?

MoriaYechiel:
Yep

PR #1123, opened 2019-05-16
Allow percent, comma, and colon in URIs

r-clancy:
These are common characters in Wikidata URIs - currently CM-Well is unable to import Wikidata dumps.

Fixes https://github.com/CM-Well/CM-Well/issues/1121

### DIFF
--- a/server/cmwell-ws/app/ld/util/LDFormatParser.scala
+++ b/server/cmwell-ws/app/ld/util/LDFormatParser.scala
@@ -86,7 +86,7 @@ object LDFormatParser extends LazyLogging {
       (45 to 57).contains(i)  ||     // (-, ., /, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9)
       (65 to 90).contains(i)  ||     // (A, B, C, D, E, F, G, H, I, J, K, L, M, N, O, P, Q, R, S, T, U, V, W, X, Y, Z)
       (97 to 122).contains(i) ||     // (a, b, c, d, e, f, g, h, i, j, k, l, m, n, o, p, q, r, s, t, u, v, w, x, y, z)
-      i == 95 || i == 35 || i > 126  // 95 = '_', 35 = '#', 123 to 126 are: ({, |, }, ~)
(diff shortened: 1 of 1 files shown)


bryaakov:
Thank you @r-clancy for your contribution!
Since this is a sensitive change, we must add some tests to make sure we are not shooting ourselves in the foot. If you would like to add integration tests in cons-it/.../RDFTests.scala and cons-it/.../SearchTests.scala that'd be nice. We can also add those if you don't have the time.

One more question: In your use case, will there be any `%,:`s in **Predicates**, or only in Subjects?

MoriaYechiel:
#1121

MoriaYechiel:
Hi @r-clancy ! thank you for your interesting and contribution to cmwell!
I saw your PR and "played" with it a little. 
1. I don't see any issue with adding comma and colon. 
2. We would like to use this PR and add also plus. I think @Ivan-Shestakov will use it
3. Percent is not "well supported" from UI and queries perspective. Maybe we should deal with it in a different issue. @bryaakov what do you think?

r-clancy:
@bryaakov I'm afraid I've got a lot on my plate right now so I won't be able to add the integration tests, at least in the next week or two. I've confirmed that the `%,:` exist in only subjects (and objects, I suppose) and not predicates.
@MoriaYechiel Would you like me to add `+` in another commit?

e-orz:
@MoriaYechiel How do comma and colon play with search? Will you need to escape the them? Otherwise they will be interpreted as a combination of two search queried for comma and equality for the colon.

MoriaYechiel:
@r-clancy I was thinking about adding + instead of % , and work on percent in another issue (since adding it is not that straight forward) WDYT ? 

@e-orz Adding those chars will affect on searching by path, which according to Yaakov is not that common, so we can start by adding these characters and then work on support qp queries on a separate issue.

r-clancy:
@MoriaYechiel Right now my use case needs `%`, for example [https://www.wikidata.org/wiki/Special:EntityData/Q76.nt](https://www.wikidata.org/wiki/Special:EntityData/Q76.nt) contains triples with `%` characters. However, if the implementation of `%` is more complex than the others than a separate issue would be fine with me.

PR #1140, opened 2019-05-29
Expand number of transport addresses

MoriaYechiel:


server/cmwell-cons/src/main/scala/ctl.scala
```diff
@@ -231,6 +231,16 @@ abstract class Host(user: String,
   def getMinMembers = minMembers.getOrElse(ips.size / 2 + 1)
   def getCasUseCommitLog = casUseCommitLog
 
+  def getThreesome(list: List[String], num: String): String = {
```
e-orz: change num

PR #1141, opened 2019-05-30
Coordinator node: Health PE, Sniff and ps validation

MoriaYechiel:
1. Fix Health monitor for PE. line "Coordinator is down" will not be shown since will anyhow not run on PE

2. Sniff set to false on WS & CW (so they "talk" only with Coordinator)

3. Validate there's no place were we call `ps` and do not consider Coordinator


server/cmwell-controller/src/main/scala/cmwell/ctrl/controllers/ElasticsearchController.scala
```diff
@@ -24,6 +24,9 @@ import scala.concurrent.ExecutionContext.Implicits.global
   */
 object ElasticsearchController
     extends ComponentController(s"${Config.cmwellHome}/app/es/cur", "/log/es[0-9]*", Set("es")) {
+
+  override val startScriptPattern: String = """-E "start[0-9]*|(-coordinator).sh""""
```
e-orz: this file shouldn't be here, and anyway it has the `()` bug in it...

PR #1143, opened 2019-06-02
dc-sync: add transformations support to monitor

e-orz:


server/cmwell-controller/src/main/scala/cmwell/ctrl/checkers/DcChecker.scala
```diff
@@ -126,9 +133,7 @@ object DcChecker extends Checker with RestarterChecker with LazyLogging {
               getLastStates(1).headOption match {
                 case Some(sb: DcStatesBag) =>
                   val lastState = sb.states.get(id)
-                  logger.debug(
-                    s"DcChecker: Last state was $lastState, remote index time: ${dcDiff.indextimeDiff.remote}"
-                  )
+                  logger.debug(s"DcChecker: Last state was $lastState, remote index time: ${dcDiff.indextimeDiff.remote}")
```
MoriaYechiel: Add id

PR #1154, opened 2019-06-12
Starting to work on traversal path

MoriaYechiel:


bryaakov:
0. Nicely done!
1. Please please please get rid of FA. Try something [like this](https://unicode-table.com/en/sets/arrows-symbols/) instead.

PR #1200, opened 2019-08-06
Support _ow

MoriaYechiel:


e-orz:
_ow must support syncing from old env that doesn't have lastModifiedBy.

PR #1212, opened 2019-09-10
add cluster config example

Ivan-Shestakov:


examples/cluster
```diff
@@ -0,0 +1,53 @@
...
+val instDirs = InstDirs("/mnt/d1/cmwell-installation", "/opt")
+
+val cas = JvmMemoryAllocations(mxmx = 2048, mxms = 2048, mxmn = 256, mxss = 256)
+val es = JvmMemoryAllocations(mxmx = 4096, mxms = 4096, mxmn = 256, mxss = 256)
+val bg = JvmMemoryAllocations(mxmx = 2048, mxms = 2048, mxmn = 256, mxss = 256)
+val ws = JvmMemoryAllocations(mxmx = 1024, mxms = 1024, mxmn = 256, mxss = 256)
+val cluster = Grid(user = "cmuser" ,
+              password = "" ,
+              clusterIps = hostnames,
+              inet = "eth0",
```
MoriaYechiel: We've removed inet . Should not be in any of files

examples/cluster
```diff
@@ -0,0 +1,53 @@
...
+// Note: a sudoer is needed for first time installation.
+// Save this file as "cluster", and once you've built CM-Well, run CM-Well Console:
+// server/cmwell-cons/app/cmwell.sh
+//
+// within CM-Well Console, type the following:
+// :load cluster
+// cluster.prepareMachines()
+// cluster.install
+
+val hostnames = (1 to 10).map("cmw" + _)
```
MoriaYechiel: Suggest to create a "stupid" list like: 
val hostnames = List("10.10.10.1", "10.10.10.2", "10.10.10.3" etc )

PR #1225, opened 2019-09-24
Fix tests to work with github CI CD

MoriaYechiel:
Grid test - global execution context gives us only 2 threads in thread pull.
BG - uses spinCheck instead of sleep


server/cmwell-bg/src/test/scala/cmwell/bg/test/BGResilienceSpec.scala
```diff
@@ -99,34 +95,32 @@ class BGResilienceSpec extends FlatSpec with BeforeAndAfterAll with BgEsCasKafka
...
+    val esCheck = cmwell.util.concurrent.spinCheck(1.second, true, 60.seconds) {
+      try{
         ftsServiceES.search(
-          pathFilter = None,
-          fieldsFilter = Some(SingleFieldFilter(Must, Equals, "system.path", Some(s"/cmt/cm/bg-test/circumvented_bg/info$i"))),
-          datesFilter = None,
-          paginationParams = PaginationParams(0, 200)
-        ),
-        10.seconds
+        pathFilter = None,
```
e-orz: You could use the pathFilter with descendants instead of manually doing it with system.parent.parent_hierarchy

PR #1238, opened 2019-10-06
fix missing uuid log

lielShraga:


server/cmwell-dc/src/main/scala/cmwell/dc/stream/InfotonRetriever.scala
```diff
@@ -53,22 +53,24 @@ object InfotonRetriever extends LazyLogging {
...
+        case ExtraData(newSubject, None, lmb @ Some(_), None) => Some(this.copy(subject = newSubject, lastModifiedBy = lmb))
+        case ExtraData(newSubject, None, None, uuid @ Some(_)) => Some(this.copy(subject = newSubject, uuid = uuid))
+
       }
     }
 
     def getQuads(indexTimeN: Long, modifierN: Option[String]) = {
       def modifier = modifierN.fold(throw ModifierMissingException(subject))(identity)
-
+      uuid.fold("")(u => s"""<$subject> <cmwell://meta/sys#uuid> "$u"^^<http://www.w3.org/2001/XMLSchema#string> .\n""") ++
```
e-orz: there should ALWAYS by uuid. You should throw an exception if not.

server/cmwell-dc/src/main/scala/cmwell/dc/stream/SingleMachineInfotonIngester.scala
```diff
@@ -59,7 +59,13 @@ object SingleMachineInfotonIngester extends LazyLogging {
         case (infotonSeq, state) => {
           val payloadBuilder = new ByteStringBuilder
           // no need for end line because each line in already suffixed with it
-          infotonSeq.foreach(payloadBuilder ++= _.data)
+          infotonSeq.foreach{i=>
+            val firstLine = i.data.takeWhile(_ != stream.newLine)
+            if(i.data.takeWhile(_ != stream.newLine).contains(ByteString("meta/sys#uuid")))
```
e-orz: 1. you create the same ByteString for each infoton - make a val for it.
2. i.data.takeWhile(_ != stream.newLine) is done twice - make a val
3. contains seems to not be working in this case

server/cmwell-dc/src/main/scala/cmwell/dc/stream/SingleMachineInfotonIngester.scala
```diff
@@ -59,7 +59,13 @@ object SingleMachineInfotonIngester extends LazyLogging {
...
+              payloadBuilder ++= i.data.take(firstLine.length)
```
e-orz: 1. Third time the same `take`.
2. If the first line is uuid take *only* the first line?

server/cmwell-dc/src/main/scala/cmwell/dc/stream/InfotonRetriever.scala
lielShraga: done

server/cmwell-dc/src/main/scala/cmwell/dc/stream/SingleMachineInfotonIngester.scala
lielShraga: 1. I compare to string so it's not relevant in the new impl
2.done
3.done

server/cmwell-dc/src/main/scala/cmwell/dc/stream/SingleMachineInfotonIngester.scala
lielShraga: 1.done. I changed it to drop
2. it's working correctly

server/cmwell-dc/src/main/scala/cmwell/dc/stream/InfotonRetriever.scala
```diff
@@ -131,7 +131,8 @@ object InfotonRetriever extends LazyLogging {
                       throw WrongPathGotException(s"Got path ${path} from _out that was not in the uuids request bulk: ${
                         state._1.map(i => i.uuid.utf8String + ":" + i.base.path).mkString(",")}")
                     case Some((builder, curExtraData)) => {
-                      builder ++= (nquad ++ endln)
+                      val finalNquad = if(nquad.utf8String.isEmpty) nquad else nquad ++ endln
```
e-orz: 1. Why not simply:
`if(nquad.utf8String.nonEmpty) builder ++= finalNquad`
It's an imperative part of the code - use normal if for it. Why adding an empty nquad?
1. you are practically converting every nquad to a string. But all you want to check is whether it's empty or not.

PR #1245, opened 2019-10-07
Warning hackathon - Util

MoriaYechiel:


server/cmwell-util/src/main/scala/cmwell/util/concurrent/SimpleScheduler.scala
```diff
@@ -59,11 +59,8 @@ object SimpleScheduler extends LazyLogging {
       }
     }
 
-    //prepare returns `this` anyway in all seen cases... (can't see why NOT do this)
-    val ec = executionContext.prepare()
-
     val cancellable = timer.scheduleAtFixedRate(new Runnable {
-      override def run(): Unit = ec.execute(runnable)
+      override def run(): Unit = executionContext.execute(runnable)
```
MoriaYechiel: https://www.scala-lang.org/api/current/scala/concurrent/ExecutionContext.html#prepare():scala.concurrent.ExecutionContext

PR #1250, opened 2019-10-07
Fix warning on cmwell-plugin-gremlin

MoriaYechiel:


server/cmwell-plugin-gremlin/src/main/scala/cmwell/plugins/impl/GremlinParser.scala
```diff
@@ -61,7 +59,12 @@ class GremlinParser extends SgEngineClient {
       typedPipe
     }
 
-    def read(p: Pipe[_, _]) = p.iterator().mkString("\n")
+    def read(p: Pipe[_, _]) = {
```
MoriaYechiel: Could that be done better?

e-orz: it works:
`p.iterator().asScala.mkString("\n")`

PR #1265, opened 2019-10-29
System Fields - Full Test

MoriaYechiel:


bryaakov:
Why do we need `<http://feed.tms/1.0/...> <cmwell://meta/sys#protocol> "http" .` in each Infoton in test data?
Isn't the protocol being parsed from Subject URI ?

PR #1266, opened 2019-11-05
Add proxy support

MoriaYechiel:


server/cmwell-data-tools/src/main/scala/cmwell/tools/data/utils/akka/AkkaUtils.scala
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  def generateConnectionPoolSettings(userAgent : Option[String] = None) ={
+    val settings = userAgent.fold{
+      ConnectionPoolSettings(config)
+      .withConnectionSettings(ClientConnectionSettings(config))
+    }{_ =>
+      ConnectionPoolSettings(
+        ConfigFactory
+          .parseString(s"akka.http.host-connection-pool.client.user-agent-header=$userAgent")
```
e-orz: `userAgent` here is an `Option`. You should leverage `fold` and extract the value using it.

PR #1267, opened 2019-11-06
New JWT library

MoriaYechiel:


bryaakov:
Looking good! Two questions:
1. Why do need uJson, when we already have PlayJson?
2. In `claimsSet`, why are we invoking `decodeRaw` twice, and hoping the second time would succeed if the first one was failing? And if that's the way to go, can't we use `Try.orElse` instead of `if(d.isSuccess) d.get` etc. ?

MoriaYechiel:
1. Cause that's a nice library :)  How "bad" is it to add it?
2. It's a mistake! will fix it. (Happened cause at first I used the secret & secret2 to parse)

BTW - wow for the response! Didn't expect it to be so fast

bryaakov:
I will kindly ask @e-orz to answer the "how \"bad\" is it" question. IMHO we need a good reason to add more libraries, but I'll go with what he says in this case.

e-orz:
There is no reason why not adding it. It has NO run time dependencies and it won't interfere with other libraries.

PR #1269, opened 2019-11-06
New Semantic Version Library

MoriaYechiel:


server/cmwell-cons/src/main/scala/Upgrade.scala
```diff
@@ -39,16 +39,28 @@ object Upgrade {
   }
 
   def runPostUpgradeActions(currentVersionF : Future[String], upgraded : String, hosts: GenSeq[String]): Future[Boolean]= {
-    def shouldRun(test : SemVersion, curr : SemVersion, upgrade : SemVersion) : Boolean = (test > curr) && (test <= upgrade)
+    /**
+      *  Note! According to semantic versions rules, what will be taken into account is only <major>.<minor>.<patch>
+      *    all test after + sign will not affect precedence!
```
e-orz: all *test*???

server/cmwell-cons/src/main/scala/Upgrade.scala
```diff
@@ -39,16 +39,28 @@ object Upgrade {
...
+      *    Examples:  1.2.5  >  1.2.5
```
e-orz: in this case it will be `==`

server/cmwell-cons/src/main/scala/Upgrade.scala
```diff
@@ -39,16 +39,28 @@ object Upgrade {
...
 
     currentVersionF.foreach{current =>
-      val currentVersion = Version(current)
-      val upgradedVersion = Version(upgraded)
+      val currentVersion = SemVer(current)
+      val upgradedVersion = SemVer(upgraded)
 
-      val relevantUpgradeFunctions = postUpgradeList.filter(f => shouldRun(Version(f.applyToVersion), currentVersion, upgradedVersion))
+      if (currentVersion.isEmpty || upgradedVersion.isEmpty)
+          error(s"Failed to parse the versions:\n ${
```
e-orz: According to your code, the "good" version will be displayed and not the bad one. It will confuse the user.

server/cmwell-cons/src/main/scala/Upgrade.scala
MoriaYechiel: OOPS! meant "text"

server/cmwell-cons/src/main/scala/Upgrade.scala
MoriaYechiel: I think it's ok.
f.e. if currentVersion is empty (which I dont expect that!), it will enter the first part of fold:
(s"current: $current")
and prints the value (current) that let to "None" in currentVersion

PR #1274, opened 2019-11-10
Add cache mechanism

MoriaYechiel:


.github/workflows/CI.yml
```diff
@@ -13,9 +13,21 @@ jobs:
...
     runs-on: ubuntu-18.04
 
-    steps:
+    steps:  
       - name: Checkout code
         uses: actions/checkout@v1
+      - name: Cache .ivy2
+        uses: actions/cache@v1
+        with:
+          path: /home/runner/.ivy2
```
e-orz: Why don't you use environments variables like `HOME` or `GITHUB_WORKSPACE`

.github/workflows/CI.yml
```diff
@@ -13,9 +13,21 @@ jobs:
...
+          key: cmwell-unitTests-ivy2-${{ hashFiles('**/build.sbt') }}
+          restore-keys: cmwell-unitTests-ivy2-
```
e-orz: According to their docs when there are multiple restore keys:
1. If there is an exact match - take the key. In your case there won't be an exact match (the key itself always includes the hash)
2. On partial match (by prefix, I think): take into account **all** the partially matched restored keys and take the most recent one.
Using a prefix like `cmwell-unitTests-ivy2-` actually "kills" the ability to use cache from integration tests. Using `cmwell-ivy2-unitTests-` (in key and in restore keys) will allow it. Do you want it?
3. How are you going to solve the mac issue what its PR will be merged (if at all)?

e-orz: Thought about it again: you can use `cmwell-unitTests-ivy2-` and add another restore keys (that will point to the other job)

.github/workflows/CI.yml
```diff
@@ -67,14 +80,38 @@ jobs:
...
       run: |
-        wget https://downloads.lightbend.com/scala/2.13.1/scala-2.13.1.tgz
-        sudo tar -xf scala-2.13.1.tgz
-        sudo ln -s `pwd`/scala-2.13.1/bin/scala /usr/local/bin/scala
+        scalaVersion="2.13.1"
+        if [ ! -d "`pwd`/scala-$scalaVersion" ]; then
+          echo "`pwd`"
+          echo "Scala does not exist! Going to download"
+          wget https://downloads.lightbend.com/scala/$scalaVersion/scala-$scalaVersion.tgz
+          sudo tar -xf scala-$scalaVersion.tgz
```
e-orz: extract always to the same place.
Don't forget to delete (rm) the directory before (why directory is called like this and not "folder" :smile: )

PR #1279, opened 2019-11-18
IntegrationTests to be ran on macOS as well

MoriaYechiel:


MoriaYechiel:
@e-orz I know you won't like the "latest", seems that it's the only available option:
https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobsjob_idruns-on

PR #1281, opened 2019-11-18
ServicesRoutes: init. impl.: Only Redirection support

bryaakov:


server/cmwell-ws/app/logic/services/ServicesRoutesCache.scala
```diff
@@ -0,0 +1,94 @@
...
+@Singleton
+class ServicesRoutesCache @Inject()(crudService: CRUDServiceFS)(implicit ec: ExecutionContext, sys: ActorSystem) extends LazyLogging {
+  private var services: Map[String, ServiceDefinition] = Map.empty
+  private val helperActor = sys.actorOf(Props(new ServicesRoutesCacheHelperActor), "ServicesRoutesCacheHelperActor")
+  private implicit val timeout: Timeout = 60.seconds
+
+  def find(path: String): Option[ServiceDefinition] =
+    services.find { case (route, _) => path.startsWith(route) }.map(_._2)
+
+  def list: Set[String] = services.keySet
```
MoriaYechiel: I think servicesList will be a better name

server/cmwell-ws/app/logic/services/ServiceDefinition.scala
```diff
@@ -0,0 +1,31 @@
...
+
+sealed abstract class LogicService(override val route: String, logicPath: String,
+                                   args: Array[String], returnType: ServiceReturnType) extends ServiceDefinition
+
+case class Source(override val route: String, logicPath: String,
+                  args: Array[String], returnType: ServiceReturnType) extends LogicService(route, logicPath, args, returnType)
+case class Binary(override val route: String, logicPath: String,
+                  args: Array[String], returnType: ServiceReturnType) extends LogicService(route, logicPath, args, returnType)
+
+case class ServiceReturnType(infotonType: String, mimeType: Option[String])
```
MoriaYechiel: We talked about commenting it out.
Maybe it's best to add something like "Feature under construction"

server/cmwell-ws/app/logic/services/ServicesRoutesCache.scala
```diff
@@ -0,0 +1,94 @@
...
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Failure, Success, Try}
+import akka.pattern.ask
+import akka.util.Timeout
+import scala.concurrent.duration.DurationInt
+
+@Singleton
+class ServicesRoutesCache @Inject()(crudService: CRUDServiceFS)(implicit ec: ExecutionContext, sys: ActorSystem) extends LazyLogging {
+  private var services: Map[String, ServiceDefinition] = Map.empty
```
MoriaYechiel: I have a problem with this map..
You change it from within the actor, but you read it from the outside. So you're not really thread-safe.
Maybe @volatile use is better? (I'm not sure!!)

server/cmwell-ws/app/logic/services/ServicesRoutesCache.scala
```diff
@@ -0,0 +1,94 @@
...
+        case Failure(t) => logger.error("Could not load Services from /meta/services", t)
+      }.map(_ => ())
+    }
+
+    override def receive: Receive = {
+      case Populate =>
+        val s = sender()
+        populate().andThen { case _ =>
+            context.system.scheduler.scheduleOnce(Settings.servicesRoutesCacheRefreshInterval, self, Populate)
+            s ! ()
```
MoriaYechiel: What's the use with replying to the sender?

server/cmwell-it/src/it/scala/cmwell/it/ServicesRoutesTests.scala
```diff
@@ -0,0 +1,67 @@
...
+import scala.concurrent.duration.DurationInt
+
+class ServicesRoutesTests extends AsyncFunSpec
+  with Matchers
+  with Inspectors
+  with OptionValues
+  with EitherValues
+  with Helpers
+  with LazyLogging {
+
```
MoriaYechiel: Are all imports needed..?

server/cmwell-ws/app/logic/services/ServicesRoutesCache.scala
bryaakov: but then the usag would be `servicesRoutesCache.servicesList` - we already know those are services, so .list feels like a better fit IMO.

server/cmwell-ws/app/logic/services/ServicesRoutesCache.scala
bryaakov: This is thread-safe enough for the purpose. In the worst case, a GET won't see the most updated picture of services, and user will have 404, in case a service was defined (i.e. ingested to /meta/services) but not yet consumed. Maybe we can lower the refresh rate from 5 minutes to 1.

MoriaYechiel: No need to change refresh rate

PR #1282, opened 2019-11-21
Stop stream only after emitting all

MoriaYechiel:


MoriaYechiel:
1. Closing the port (out) before emitting all, will make us loose them. 
2. Not closing the port, will leave the stream running

e-orz:
Using ` while (!buf.isEmpty) {` will cause sending elements downstream even though non requested.
Use emitMultiple instead.

PR #1303, opened 2020-01-15
STP Quick Fix

YuriiYarosh:
Quick fix for sort by parsing.

SparqlProcessorManager needs a [proper retry](https://github.com/CM-Well/CM-Well/blob/master/server/cmwell-sparql-agent/src/main/scala/cmwell/tools/data/sparql/SparqlProcessorManager.scala#L608) for 400 statuses, because right now retrying 503 with a delay seams to work fine, but 400'ies result in a short-circuit.

I've added a [special case](https://github.com/CM-Well/CM-Well/pull/1303/commits/865e61416b2c6f64c8bdbd9b6d57e7e3a798bf59#diff-e3aab043a3f445ff9ba6a445bcc50908) for that, but that's far from optimal.

Needs refactoring.

### DIFF
--- a/server/cmwell-data-tools/src/main/scala/cmwell/tools/data/utils/akka/Retry.scala
+++ b/server/cmwell-data-tools/src/main/scala/cmwell/tools/data/utils/akka/Retry.scala
@@ -177,13 +177,20 @@ object Retry extends DataToolsLogging with DataToolsConfig {
           }
           else {
             // server error
-            // special case: sparql-processor //todo: remove this in future
+            // special case: sparql-processor //TODO: remove this in future
 
             e.dataBytes.runFold(ByteString.empty)(_ ++ _).map(_.utf8String).map{ entity=>
               logger.warn(s"[$errorID] server error. Body of response: $entity, headers: ${headersString(h)}")
             }
 
-            if (e.toString contains "Fetching") {
+            // special case: sparql-processor //TODO: remove this in future
+            if (e.toString contains "SpHandler Parsing error") {
+              logger.warn(
+                s"$labelValue will not schedule a retry for the SpHandler Parsing Error"
+              )
+
+              None
+            } else if (e.toString contains "Fetching") {
               logger.warn(
                 s"$labelValue will not schedule a retry on data ${concatByteStrings(data, endl).utf8String}"
               )
--- a/server/cmwell-dc/src/main/scala/cmwell/dc/stream/TsvRetriever.scala
+++ b/server/cmwell-dc/src/main/scala/cmwell/dc/stream/TsvRetriever.scala
@@ -183,9 +183,11 @@ object TsvRetriever extends LazyLogging {
     val (host, port) = hostPort.head -> hostPort.tail.headOption
       .getOrElse("80")
       .toInt
+
     val tsvPoolConfig = ConfigFactory
-      .parseString("akka.http.host-connection-pool.max-connections=1")
+      .parseString("akka.http.host-connection-pool.max-connections=20")
       .withFallback(config)
+
     val tsvConnPool = Http()
       .newHostConnectionPool[TsvRetrieveState](
         host,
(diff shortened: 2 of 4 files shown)


bryaakov:
Thank you Yurii for your PR! It's great to have you here. Please see some inline comments.

server/cmwell-data-tools/src/main/scala/cmwell/tools/data/utils/akka/Retry.scala
```diff
@@ -177,13 +177,20 @@ object Retry extends DataToolsLogging with DataToolsConfig {
...
             // server error
-            // special case: sparql-processor //todo: remove this in future
+            // special case: sparql-processor //TODO: remove this in future
 
             e.dataBytes.runFold(ByteString.empty)(_ ++ _).map(_.utf8String).map{ entity=>
               logger.warn(s"[$errorID] server error. Body of response: $entity, headers: ${headersString(h)}")
             }
 
-            if (e.toString contains "Fetching") {
+            // special case: sparql-processor //TODO: remove this in future
```
bryaakov: this is a Client error (400) and should be in a different case (line 231)

server/cmwell-dc/src/main/scala/cmwell/dc/stream/TsvRetriever.scala
```diff
@@ -183,9 +183,11 @@ object TsvRetriever extends LazyLogging {
     val (host, port) = hostPort.head -> hostPort.tail.headOption
       .getOrElse("80")
       .toInt
+
     val tsvPoolConfig = ConfigFactory
-      .parseString("akka.http.host-connection-pool.max-connections=1")
+      .parseString("akka.http.host-connection-pool.max-connections=20")
```
bryaakov: This is TsvRetriever of **DC Sync** component - unrelated to STP

server/cmwell-ws/app/controllers/Application.scala
```diff
@@ -2139,7 +2141,9 @@ callback=< [URL] >
         val xg = queryString.keySet("xg")
         val yg = queryString.keySet("yg")
 
-        if (offset > Settings.maxOffset) {
+        if (rawSortParamsTry.isFailure) {
```
bryaakov: Nice fix! But I think it deserves its own PR. I don't see how it is related to STP fix.

server/cmwell-ws/app/controllers/SpHandler.scala
```diff
@@ -89,7 +89,7 @@ class SpHandlerController @Inject()(crudServiceFS: CRUDServiceFS)(implicit ec: E
     import SpHandler._
 
     Try(parse(req)) match {
-      case Failure(_) => Future.successful(wsutil.exceptionToResponse(new IllegalArgumentException("Parsing error")))
+      case Failure(_) => Future.successful(wsutil.exceptionToResponse(new IllegalArgumentException("SpHandler Parsing error")))
```
bryaakov: Exposing internal names of classes as an Exception that the user see (it will be converted to an HTTP 400) is not a good practice in terms of security. I'd suggest not to change "SpHandler Parsing error" to "_sp API: parsing error" or perhaps just leave it as it is today (I know you're matching against it in STP, but why not match to "Parsing error"?)

server/cmwell-data-tools/src/main/scala/cmwell/tools/data/utils/akka/Retry.scala
YuriiYarosh: Yes, it retries requests to SpHandler over and over which returns 400 Parsing Error.

server/cmwell-ws/app/controllers/Application.scala
YuriiYarosh: Yes it's unrelated.
Just got fixed along the way.

server/cmwell-ws/app/controllers/SpHandler.scala
YuriiYarosh: > or perhaps just leave it as it is today

I'll work on proper retry mechanism later - it needs a lot of attention.
[shouldRetry](https://github.com/CM-Well/CM-Well/blob/master/server/cmwell-sparql-agent/src/main/scala/cmwell/tools/data/sparql/SparqlProcessorManager.scala#L608) was commented in SparqlProcessorManager.

I could refactor SparqlProcessorManager shouldRetry and Retry later - i don't think it's a priority right now.

YuriiYarosh:
I'll prep a better one.

PR #1306, opened 2020-01-22
logback sanitizing encoder and sanitizing string interpolator

YuriiYarosh:
Related to [#1028](https://github.com/CM-Well/CM-Well/pull/1028) _CRLF Injection Protection by Sanitization, as recommended by VeraCode_

Created another pull request on top of it.

Nothing major 
1) Added a `san` sanitizing string interpolator [over here](https://github.com/YuriiYarosh/CM-Well/blob/vera/server/cmwell-util/src/main/scala/cmwell/util/string/package.scala#L336)
2) A logback sanitizing pattern encoder [here](https://github.com/YuriiYarosh/CM-Well/blob/vera/server/cmwell-util/src/main/scala/cmwell/util/log/SanitizingPatternEncoder.scala#L23)
3) And a small [test case](https://github.com/YuriiYarosh/CM-Well/blob/vera/server/cmwell-util/src/test/scala/cmwell/util/string/test/StringTests.scala#L65)


YuriiYarosh:
IDEA automatically picked up current scalafmt config ... so, a lot of noise from the scalafmt.

Should we make a reformat PR alongside ?

YuriiYarosh:
I could discard all the formatting amends if needed.

e-orz:
Yes, please discard the formatting. We want each PR to handle only one thing. Thanks.

PR #1310, opened 2020-01-28
removed old CMWellBuild sbt plugins, added bloop config

YuriiYarosh:
So, I've removed obsolete sbt plugins and added a simple bloop config.

```bash
brew install scalacenter/bloop/bloop
sbt bloopInstall # creates a bloop config and exports sbt projects
bloop server     # in a separate shell or screen
bloop compile server
bloop test server --propagate 
```

It's been tested against Scala 2.13.1


YuriiYarosh:
Sure, idea has't picked that up.

PR #1317, opened 2020-02-10
simplifing sbt structure and removing unused sbt plugins

e-orz:
The is the same as PR #1311 that @YuriiYarosh did.
I only fixed indentations and reordered the functions so Git will take the file as rename and not as new and deleted.


server/project/CMWellBuild.scala
```diff
@@ -230,24 +238,35 @@ object CMWellBuild extends AutoPlugin {
...
+	}
+
 
-	override def projectSettings = Seq(
-		scalastyleFailOnError := false,
-		testScalastyle in ThisProject := (scalastyle in ThisProject).in(Test).toTask("").value,
-		(test in Test) := ((test in Test) dependsOn testScalastyle).value,
-		compileScalastyle in ThisProject := (scalastyle in ThisProject).in(Compile).toTask("").value,
-		(compile in Compile) := ((compile in Compile) dependsOn compileScalastyle).value,
+	def projectSettings = Seq(
```
MoriaYechiel: Change to val

PR #1320, opened 2020-03-12
subjectsInSpAreHttps is now list of paths instead of boolean

bryaakov:


server/cmwell-cons/src/main/scala/ComponentConf.scala
```diff
@@ -646,7 +646,7 @@ case class CwConf(home: String,
           "-Dcom.sun.management.jmxremote.ssl=false",
           "-Dcom.sun.management.jmxremote.authenticate=false",
           "-Duser.timezone=GMT0",
-          s"-Dcrashableworker.subjectsAreHttps=$subjectsInSpAreHttps"
+          s"-Dcrashableworker.subjectsAreHttps=${subjectsInSpAreHttps.mkString(",")}"
```
e-orz: remove this line. no -Ds please....

Use HOCON and create arrays and then `config.getStringList`
In order to programmatically configure it use the `{{}}` of mkConf

server/cmwell-ws/app/Settings.scala
```diff
@@ -120,7 +120,7 @@ object Settings {
 
   lazy val maxQueryResultsLength = config.getInt("crashableworker.results.maxLength")
   lazy val queryResultsTempFileBaseName = config.getString("crashableworker.results.baseFileName")
-  lazy val subjectsInSpAreHttps = config.getBoolean("crashableworker.subjectsAreHttps")
+  lazy val subjectsInSpAreHttps = config.getString("crashableworker.subjectsAreHttps")
```
e-orz: same as before - use proper List

server/cmwell-ws/src/pack/application.conf
```diff
@@ -152,7 +152,7 @@ crashableworker {
     maxLength = 1400000
     baseFileName = "tempFile"
   }
-  subjectsAreHttps = false
+  subjectsAreHttps = ""
```
e-orz: Should it be also with {{}}?

PR #1328, opened 2020-06-22
Not reusing same Digest Instance

bryaakov:


bryaakov:
not exactly. `msgDigester` is private, and what was in use was two others `private val`s that was only calculated once...

bryaakov:
or maybe I got it wrong? Hold on a sec...

bryaakov:
You are right, those vals are actually functions. I'm closing this PR

PR #1337, opened 2023-06-20
Update Sbt and plugins, fix some deprecations, support Java 11

lolgab:


e-orz:
Hi,
thank you for your PR.
Is there any reason you had to do those changes (updating libraries is always welcome, but still)? Didn't it work with some newer environments?

Thanks!

lolgab:
Hi @e-orz,
No, I didn't need the update (I think I updated only artifacts using in the build but not in the production code).
I found your project today and I wanted to give back updating the build to the latest sbt version.
I have only Java 11 installed in my computer, so I took the opportunity to use APIs that are supported also in Java 11 so I could compile the code.
Can you approve the Github Action pipeline?

